PR #41, opened 2017-12-19
Add an optional --extra_artifacts argument

tomchop:
Use case: we want default artifacts + BrowserHistory (which is not included by default)

### DIFF
--- a/dftimewolf/cli/recipes/grr_artifact_hosts.py
+++ b/dftimewolf/cli/recipes/grr_artifact_hosts.py
@@ -50,7 +50,7 @@ args = [
     ('reason', 'Reason for collection', None),
     ('--artifact_list', 'Comma-separated list of artifacts to fetch '
      '(override default artifacts)', None),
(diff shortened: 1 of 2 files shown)


dftimewolf/cli/recipes/grr_artifact_hosts.py
```diff
@@ -44,16 +45,12 @@
...
-    'AllShellConfigs')
-
 args = [
     ('hosts', 'Comma-separated list of hosts to process', None),
     ('reason', 'Reason for collection', None),
-    ('--artifact_list', 'Comma-separated list of artifacts to fetch',
-     DEFAULT_ARTIFACT_LIST),
+    ('--artifact_list', 'Comma-separated list of artifacts to fetch '
+     '(override default artifacts)', None),
+    ('--extra_artifacts', 'Comma-separated list of artifacts to fetch '
```
berggren: Maybe: "Comma-separated list of artifacts to append to the default artifact list"?

PR #52, opened 2018-01-15
Add SIGINT capabilities.

tomchop:


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -55,6 +56,8 @@
 from dftimewolf.cli.recipes import local_plaso
 from dftimewolf.cli.recipes import plaso_file_timesketch
 
+signal.signal(signal.SIGINT, dftw_utils.signal_handler)
```
berggren: So many signal.. :)

To get less signal, maybe do:
```
from signal import signal
```

tomchop: Sounds good. Done.

PR #59, opened 2018-04-25
Fix a bug in the hunt approval checking flow

tomchop:


dftimewolf/lib/collectors/grr_base.py
```diff
@@ -45,18 +46,29 @@ def setup(self, reason, grr_server_url, grr_auth, approvers=None):
...
     self.reason = reason
 
-  def _check_approval_wrapper(self, grr_object, function):
-    """Wraps a call to GRR functions checking for approval."""
+  def _check_approval_wrapper(self, grr_object, function, *args, **kwargs):
+    """Wraps a call to GRR functions checking for approval.
+
+    Args:
+      grr_object: the GRR object to create the eventual approval on.
+      function: The GRR function requiring approval.
```
berggren: I would call this grr_function to be more clear.
(I have mixed feelings about throwing around function pointers like this and then execute them :)

dftimewolf/lib/collectors/grr_base.py
```diff
@@ -45,18 +46,29 @@ def setup(self, reason, grr_server_url, grr_auth, approvers=None):
     self.output_path = tempfile.mkdtemp()
     self.reason = reason
 
-  def _check_approval_wrapper(self, grr_object, function):
-    """Wraps a call to GRR functions checking for approval."""
+  def _check_approval_wrapper(self, grr_object, function, *args, **kwargs):
+    """Wraps a call to GRR functions checking for approval.
+
+    Args:
+      grr_object: the GRR object to create the eventual approval on.
```
berggren: Can't we skip passing the whole object here, and just have a grr_id parameter? Then it is up to the caller to provide the correct ID.

PR #81, opened 2018-06-18
Move documentation to docs/

tomchop:
Once this is merged we can point dftimewolf.readthedocs.io to it


docs/conf.py
```diff
@@ -0,0 +1,225 @@
...
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+import re
```
Onager: Please import unicode_literals here, unless there's some reason not to do so.

docs/module-list.md
```diff
@@ -0,0 +1,44 @@
...
+## Collectors
+
+* `FilesystemCollector` - a simple collector that just passes a local path on to
+the processors.
+
+### GRR hunts
+
+Launch or fetch results from fleet-wide GRR hunts.
+  * `GRRHuntArtifactCollector` - Launches a fleet-wide GRR
+  `ArtifactCollectorFlow`
```
Onager: I think you forgot to add a description here

tomchop: I meant to write "Launches a fleet-wide GRR `ArtifactCollectorFlow`" on the same line.

docs/conf.py
tomchop: Done.

PR #83, opened 2018-06-22
Update Travis, tox and codecov configurations

Onager:
Updated Travis, tox and codecov configurations


requirements.txt
```diff
@@ -0,0 +1,4 @@
+beautifulsoup4
+grr-api-client
+pytz
+requests
```
tomchop: Did you compile this manually or did you generate it using pipenv? (I think `pipenv lock --requirements`)

PR #85, opened 2018-06-26
Bugfix while checking approval

tomchop:


dftimewolf/lib/collectors/grr_base.py
```diff
@@ -61,13 +61,17 @@ def _check_approval_wrapper(self, grr_object, grr_function, *args, **kwargs):
...
+      object_id = getattr(grr_object, 'flow_id', None)
+    if not object_id:
+      object_id = getattr(grr_object, 'client_id', None)
 
     while True:
       try:
         return grr_function(*args, **kwargs)
-      except grr_errors.AccessForbiddenError:
-        print('{0:s}: no valid approval found'.format(object_id))
+      except grr_errors.AccessForbiddenError as err:
```
Onager: Please use "as exception", per the style guide.

dftimewolf/lib/collectors/grr_base.py
```diff
@@ -61,13 +61,17 @@ def _check_approval_wrapper(self, grr_object, grr_function, *args, **kwargs):
...
+        print '{0:s}: no valid approval found: {1:s}'.format(object_id, err)
```
Onager: Please fix this, and don't use the print builtin in any new code.

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -46,8 +46,14 @@ def _get_client_by_hostname(self, hostname):
       RuntimeError: if no client ID found for hostname.
     """
     # Search for the hostname in GRR
-    print('Searching for client: {0:s}'.format(hostname))
-    search_result = self.grr_api.SearchClients(hostname)
+    print 'Searching for client: {0:s}'.format(hostname)
```
Onager: Please don't use the print function in any new code

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -147,11 +154,11 @@ def _await_flow(self, client, flow_id):
         status = client.Flow(flow_id).Get().data
       except grr_errors.UnknownError:
         msg = 'Unable to stat flow {0:s} for host {1:s}'.format(
-            flow_id, self.host)
+            flow_id, client.data.os_info.fqdn.lower())
```
Onager: Is client.data.os_info.fqdn always set? I suspect probably not, so we'll get another exception here. Consider refactoring this.

dftimewolf/lib/collectors/grr_base.py
```diff
@@ -61,13 +61,17 @@ def _check_approval_wrapper(self, grr_object, grr_function, *args, **kwargs):
       The return value of the execution of grr_function(*args, **kwargs).
     """
     approval_sent = False
-    object_id = getattr(grr_object, 'hunt_id') or getattr(grr_object, 'flow_id')
+    object_id = getattr(grr_object, 'hunt_id', None)
```
Onager: Can you check the type of grr_object here instead? This looks pretty error prone. This would also let you raise if you can't get an object ID.

dftimewolf/lib/collectors/grr_base.py
tomchop: Changed all other `except` clauses to match this

dftimewolf/lib/collectors/grr_base.py
tomchop: object_id is just used to display an error message. This seems too complex for very little added value, so I'll just remove it.

dftimewolf/lib/collectors/grr_hosts.py
tomchop: There's a lot of logic assuming valid fqdns spread around the code. I think this might be worth revisiting in a future PR.

PR #88, opened 2018-07-10
Change the way GRR hunt zipfiles are unzipped

tomchop:


dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -233,6 +234,17 @@ def _get_and_write_archive(self, hunt, output_file_path):
     hunt_archive = hunt.GetFilesArchive()
     hunt_archive.WriteToFile(output_file_path)
 
+  def _get_client_fqdn(self, client_info_contents):
+    """Extracts a GRR client's FQDN from its client_info.yaml file.
+
+    Args:
+      client_info_contents: The contents of the client_info.yaml file.
```
Onager: +returns

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -247,20 +259,34 @@ def _extract_hunt_results(self, output_file_path):
...
+            hunt_dir = f.filename.split('/')[0]
+
+          # client_info.yaml contains information used to translate client ID
+          # to FQDN; extract it and save for later
+          if f.filename.split('/')[-1] == 'client_info.yaml':
+            client_id, fqdn = self._get_client_fqdn(archive.read(f))
+            client_id_fqdn[client_id] = fqdn
+            continue
+
+          # Otherwise, extract the files for each client ID in their own
```
Onager: Otherwise compared to what? Consider removing this comment, it seems more confusing that explanatory.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -247,20 +259,34 @@ def _extract_hunt_results(self, output_file_path):
     # Extract items from archive by host for processing
     collection_paths = []
     clients = set()
+    client_id_fqdn = {}
```
Onager: Please rename this to something more explanatory - fqdn_by_client_id or client_id_to_fqdn or similar.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -247,20 +259,34 @@ def _extract_hunt_results(self, output_file_path):
     # Extract items from archive by host for processing
     collection_paths = []
     clients = set()
```
Onager: It looks like this should actually be client_ids, please rename.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -247,20 +259,34 @@ def _extract_hunt_results(self, output_file_path):
...
     try:
       with zipfile.ZipFile(output_file_path) as archive:
         items = archive.infolist()
         for f in items:
+
+          if not hunt_dir:
+            hunt_dir = f.filename.split('/')[0]
+
+          # client_info.yaml contains information used to translate client ID
+          # to FQDN; extract it and save for later
```
Onager: Please add a comment here that this file is not extracted, but skipped.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -282,7 +308,14 @@ def _extract_hunt_results(self, output_file_path):
       print('Output path {0:s} could not be removed: {1:s}'.format(
           output_file_path, exception))
 
-    return collection_paths
+    # Translate GRR client IDs to FQDNs with the information retrieved
+    # earlier
+    fqdn_collection_paths = []
+    for client_id, path in collection_paths:
+      fqdn = client_id_fqdn[client_id]
```
Onager: Consider handling a KeyError here, otherwise this is going to blow up if we ever get a client with a missing FQDN, or missing client_info. Maybe client_id_fqdn.get(client_id, client_id), so you fail back to the ID if there's a problem.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -247,20 +259,34 @@ def _extract_hunt_results(self, output_file_path):
...
-            if not os.path.isdir(client_directory):
-              os.makedirs(client_directory)
             if client_id not in clients:
+              client_directory = os.path.join(self.output_path,
+                                              hunt_dir, client_id)
               collection_paths.append((client_id, client_directory))
               clients.add(client_id)
             try:
-              archive.extract(f, client_directory)
+              archive.extract(f, self.output_path)
```
Onager: Why append a different path (L286) than you extract to?

tomchop: We want to append one directory per client so that this results in multiple timelines when sent to TimesketchExporter, but we want to extract everything in the same directory so as to preserve the GRR archive structure (and end up with something like `/tmp/<tmpdir>/<hunt_id>/<client_id>`). Extracting to an individual directory would've resulted in a directory like: `/tmp/<tmpdir>/<client_id>/<hunt_id>/<client_id>`.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: Done.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: Done.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: Done.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: Done.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: Done, removed the second comment and changed the first one a bit.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: Done.

tomchop:
@Onager PTAL

PR #89, opened 2018-07-31
Initial import of Turbinia integration

aarontp:


dftimewolf/lib/processors/turbinia.py
```diff
@@ -0,0 +1,153 @@
...
+from turbinia import TurbiniaException
+from turbinia.message import TurbiniaRequest
+
+
+class TurbiniaProcessor(BaseModule):
+  """Process cloud disks with remote Turbinia instance.
+
+  Attributes:
+    client: A TurbiniaClient object
+    disk_name: Name of the disk to process
```
tomchop: `disk_name (string)` I presume?

dftimewolf/lib/processors/turbinia.py
```diff
@@ -0,0 +1,153 @@
...
+    instance (string): The name of the Turbinia instance
+    project: The project containing the disk to process
```
tomchop: `project (string)`?

dftimewolf/lib/processors/turbinia.py
```diff
@@ -0,0 +1,153 @@
...
+    region (string): The region Turbinia is in
+    zone: The zone containing the disk to process
```
tomchop: Are `region` and `zone` two different things? If you mean "cloud zone" as in `us-central1-c` for both, then I'd suggest renaming this to `turbinia_zone (string)` and `target_zone (string)`.

dftimewolf/lib/processors/turbinia.py
```diff
@@ -0,0 +1,153 @@
...
+    _output_path: The path to output files
```
tomchop: Since this one is private (i.e. leading `_`) I'm not sure it belongs in the docstring.

dftimewolf/lib/processors/turbinia.py
```diff
@@ -0,0 +1,153 @@
...
+          'disk_name, project or zone are not all specified, bailing out',
+          critical=True)
+      return
+    self.disk_name = disk_name
+    self.project = project
+    self.zone = zone
+    self._output_path = tempfile.mkdtemp()
+
+    try:
+      config.LoadConfig()
```
tomchop: I would import this as `turbinia_config` since we already have a `config` module in dftimewolf and it may be confusing.

dftimewolf/lib/processors/turbinia.py
```diff
@@ -0,0 +1,153 @@
...
+      self.region = config.TURBINIA_REGION
+      self.instance = config.PUBSUB_TOPIC
+      if config.PROJECT != self.project:
```
tomchop: If we bail when these are different, is there a reason we don't just force `self.project` to be `config.PROJECT` besides better user experience? In that case, maybe we can do the check earlier.

dftimewolf/lib/processors/turbinia.py
aarontp: Done.

dftimewolf/lib/processors/turbinia.py
aarontp: So, they are two different things.  I renamed them to turbinia_zone and turbinia_region.

dftimewolf/lib/processors/turbinia.py
aarontp: I'll defer to you since I'm sure you want it to be consistent with the rest of the project, but I personally always prefer to keep the pseudo-private members as part of the docstrings so that they are still documented somewhere.

dftimewolf/lib/processors/turbinia.py
aarontp: Done.

dftimewolf/lib/processors/turbinia.py
aarontp: Based on our earlier conversation, I think we should make the user explicitly set the project.  If the user specifies the project and it doesn't match the turbinia project, the processing will fail anyway.

I could move the check up a little bit, but I still want to have the try/except around the config loading and we don't have the project available in init.  LMK if you have suggestions for where to move it to.   I think the only consequential thing that happens before this is the mkdtemp(), so I moved that until after the check.

dftimewolf/lib/processors/turbinia.py
tomchop: SGTM

aarontp:
Ok, PTAL.  I updated some of the recipe flags.  I tested both recipes with Turbinia HEAD and everything seemed to work.

Once https://github.com/google/turbinia/pull/265 and https://github.com/google/turbinia/pull/266 are submitted, I'll push a new Turbinia package to pypi, and then I think we can submit this.  Yay!

aarontp:
Cool, AFAICT, everything should be good now.  Tests are passing, and I did some end-to-end tests that worked successfully for both recipes.  PTAL, thanks!

PR #91, opened 2018-08-02
fix tests

tomchop:


tomchop:
Travis build fails because of the pylint tests:

```
W: 46,35: Unused argument 'unused_mock_extract' (unused-argument)
```

The regex in .pylintrc seems appropriate: `dummy-variables-rgx=_|unused`

PR #94, opened 2018-08-31
Add verify parameter to GRR modules

tomchop:
Useful when testing.

Also updated and pinned dependencies for `grr-api-client` since this feature is relatively new.

### DIFF
--- a/tests/lib/collectors/grr_base.py
+++ b/tests/lib/collectors/grr_base.py
@@ -48,9 +48,10 @@ class GRRBaseModuleTest(unittest.TestCase):
     grr_base_module.setup('random reason',
                           'http://fake/endpoint',
                           ('admin', 'admin'),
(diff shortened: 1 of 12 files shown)


tests/lib/collectors/grr_base.py
```diff
@@ -62,7 +65,8 @@ def testApprovalWrapper(self):
     grr_base_module.setup('random reason',
                           'http://fake/endpoint',
                           ('admin', 'admin'),
-                          'approver1@google.com,approver2@google.com')
+                          'approver1@google.com,approver2@google.com',
```
berggren: Not important, but would change these to approver1@example.com ..

tomchop: done.

tomchop:
Travis is failling due to something wrong in pipenv. Merging this.

PR #101, opened 2018-09-04
Better Timesketch and GRR error messages

tomchop:


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -156,15 +156,18 @@ def _await_flow(self, client, flow_id):
...
             'Unable to stat flow {0:s} for host {1:s}'.format(
                 flow_id, client.data.os_info.fqdn.lower()))
 
       if status.state == flows_pb2.FlowContext.ERROR:
         # TODO(jbn): If one artifact fails, what happens? Test.
-        raise RuntimeError(
-            '{0:s}: FAILED! Backtrace from GRR:\n\n{1:s}'.format(
-                flow_id, status.context.backtrace))
+        message = status.context.backtrace
+        if 'ArtifactNotRegisteredError' in status.context.backtrace:
```
tomchop: AFAIK all GRR provides is a string, so we unfortunately need to search through it to get more details on the error.

dftimewolf/lib/errors.py
```diff
@@ -3,12 +3,8 @@
 class DFTimewolfError(Exception):
   """Class to represent a DFTimewolfError."""
 
-  fatal = False
   message = 'An error occurred.'
 
-  def __init__(self, message, fatal=False):
+  def __init__(self, message):
```
Onager: Remove, this isn't necessary any more.

PR #102, opened 2018-09-07
Add new grr_fetch_files recipe

tomchop:


dftimewolf/cli/recipes/grr_fetch_files.py
```diff
@@ -0,0 +1,42 @@
...
+    'short_description': _short_description,
+    'modules': [{
+        'name': 'GRRFileCollector',
+        'args': {
+            'hosts': '@hosts',
+            'reason': '@reason',
+            'grr_server_url': '@grr_server_url',
+            'grr_username': '@grr_username',
+            'grr_password': '@grr_password',
+            'files': '@files',
```
Onager: These are full paths to files, right? Do you need TSK:/ prefixes? Something else?

dftimewolf/cli/recipes/grr_fetch_files.py
```diff
@@ -0,0 +1,42 @@
+"""Fetches specific files from a GRR host."""
+from __future__ import unicode_literals
+
+_short_description = 'Fetches specific files from a GRR host.'
```
Onager: This works for fetching files from multiple hosts too, please fix

dftimewolf/cli/recipes/grr_fetch_files.py
tomchop: No need. I added a clarification about this parameter supporting GRR variable interpolation.

dftimewolf/cli/recipes/grr_fetch_files.py
tomchop: Done.

PR #105, opened 2018-09-21
Thread GRR collectors (closes #106)

### LINKED ISSUE #106: Thread GRR host collectors

tomchop:


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -316,52 +317,62 @@ def setup(self,
...
+      os.makedirs(client_dir)
+    print('System type: {0:s}'.format(system_type))
+
+    # If the list is supplied by the user via a flag, honor that.
+    artifact_list = []
+    if self.artifacts:
+      print('Artifacts to be collected: {0:s}'.format(self.artifacts))
+      artifact_list = self.artifacts
+    else:
+      default_artifacts = self.artifact_registry.get(system_type, None)
```
Onager: Handle KeyError here instead of defaulting to None, there's no point trying to collect the "None" artifact, and the print() below will raise in Python 3.

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -408,35 +419,42 @@ def setup(self,
...
-    Returns:
-      list of tuples containing:
-          str: human-readable description of the source of the collection. For
-              example, the name of the source host.
-          str: path to the collected data.
-
     Raises:
       DFTimewolfError: if no files specified.
     """
     # TODO(tomchop): Thread this
```
Onager: Don't need this comment any more ;)

dftimewolf/lib/collectors/grr_hosts.py
tomchop: I'm keeping the `None` here and adding a condition to the statements below.

dftimewolf/lib/collectors/grr_hosts.py
tomchop: Indeed.

PR #108, opened 2018-09-28
Triager

pstirparo:
Created new recipe, activity_triage, that collects artifacts via GRR and then process them to look for specific keywords. The recipe is intended to be called like the following

$ dftimewolf activity_triage tomchop.greendale.xyz,admin.greendale.xyz --artifacts BrowserHistory --keywords foo|bar|meow

An should print to stdout as output, one line per each "file:keyword" pair match, i.e. only one line even if the same keyword appears multiple times on the same file:
/path/to/file.ext:keyword1
/path/to/file.ext:keyword2
/path/to/anotherfile.ext:keyword1


dftimewolf/cli/recipes/activity_triage.py
```diff
@@ -0,0 +1,52 @@
...
+- Process them locally with grep to extract keywords
+"""
+from __future__ import unicode_literals
+
+_short_description = ('Fetches default artifacts from a list of GRR hosts '
+                      'and processes them with grep.')
+
+contents = {
+    'name':
+        'insider_triage',
```
Onager: Can we call this "artifact_grep" instead, to be a bit more descriptive? Also, GRR can do a grep remotely, using a condition on a file finder flow. It might need some tweaking on the GRR side to do a grep on artifact files though.

pstirparo: I've changed to "activity_triage" (like the file name, etc.), I had missed this one bit still. thanks.
I would keep the "triage" part as this recipe is going to be expanded ideally. The goal is that you, the analyst, want to have a quick glimpse on whether there has been activity regarding specific topics/keywords

dftimewolf/cli/recipes/activity_triage.py
```diff
@@ -0,0 +1,52 @@
...
+- Process them locally with grep to extract keywords
+"""
+from __future__ import unicode_literals
+
+_short_description = ('Fetches requested artifacts from a list of GRR hosts '
+                      'and processes them with grep.')
+
+contents = {
+    'name':
+        'activity_triage',
```
Onager: activity_triage is even more generic ;) What's wrong with artifact_download_grep, or something similar? That's what in the description....

pstirparo: Done. Changed to "artifact_grep"

Onager:
Ping for @pstirparo - this is dup of https://github.com/log2timeline/dftimewolf/pull/122 and can be closed, right?

pstirparo:
Yes, I had left it still open to move the files to the clean PR.

PR #110, opened 2018-10-04
Updated pylint configuration, updated dependency related files.

Onager:


tox.ini
```diff
@@ -8,6 +8,7 @@ setenv =
 deps =
     mock
     pytest
+    -rrequirements.txt
```
tomchop: What syntax is this? It kinda looks like a typo

Onager: It runs pip to install the deps in requirements.txt

PR #112, opened 2018-10-09
Python 3 compatibility

tomchop:


tomchop:
Installation of grr-api-client is hacky since it is not Python 3 compatible yet. This prevents Pipenv from correctly generating a Pipfile.lock

PR #122, opened 2018-11-06
add artifact_grep recipe

pstirparo:
Created new recipe, artifact_grep, that collects artifacts via GRR and then process them to look for specific keywords. The recipe is intended to be called like the following

$ dftimewolf artifact_grep tomchop.greendale.xyz,admin.greendale.xyz --artifacts BrowserHistory --keywords foo|bar|meow

And should print to stdout as output, one line per each "file:keyword" pair match, i.e. only one line even if the same keyword appears multiple times on the same file:
/path/to/file.ext:keyword1,keword3
/path/to/file.ext:keyword2
/path/to/anotherfile.ext:keyword1


dftimewolf/cli/recipes/artifact_grep.py
```diff
@@ -0,0 +1,52 @@
+"""Collect artifacts from hosts using GRR.
+
+- Collect a predefined list of artifacts from hosts using GRR
+- Process them locally with grep to extract keywords
+"""
+from __future__ import unicode_literals
+
+_short_description = ('Fetches requested artifacts from a list of GRR hosts '
+                      'and processes them with grep.')
```
tomchop: I'd try to be a bit more explicit here, such as "runs grep with a list of keywords on them"

dftimewolf/cli/recipes/artifact_grep.py
```diff
@@ -0,0 +1,52 @@
...
+     '(override default artifacts)', None),
+    ('--extra_artifacts', 'Comma-separated list of artifacts to append '
+     'to the default artifact list', None),
+    ('--use_tsk', 'Use TSK to fetch artifacts', False),
+    ('--approvers', 'Emails for GRR approval request', None),
+    ('--grr_server_url', 'GRR endpoint', 'http://localhost:8000'),
+    ('--verify', 'Whether to verify the GRR TLS certificate', True),
+    ('--grr_username', 'GRR username', 'admin'),
+    ('--grr_password', 'GRR password', 'admin'),
+    ('--keywords', 'Pipe separated list of keywords to search for '
```
tomchop:
```suggestion
    ('--keywords', 'Pipe-separated list of keywords to search for '
```

Also are these keywords or regular expressions?

Also, would it make sense to provide an optional parameter to a file containing a list of keywords to use? (maybe in a future PR)

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+    for _, path in self.state.input:
+      log_file_path = os.path.join(self._output_path, 'grepper.log')
+      print('Log file: {0:s}'.format(log_file_path))
+
+      print('Walking through dir (absolute) = ' + os.path.abspath(path))
+      try:
+        for root, _, files in os.walk(path):
+          for filename in files:
+            found = set()
+            fullpath = os.path.abspath(root) + '/' + filename
```
tomchop:
```suggestion
            fullpath = '{0:s}/{1:s}'.format(os.path.abspath(root), filename)
```

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+            if mimetypes.guess_type(filename)[0] == 'application/pdf':
+              found = self.grepPDF(fullpath)
+            else:
+              with open(fullpath, 'r') as fp:
+                for line in fp:
+                  found.update(set(x.lower() for x in re.findall(
+                      self._keywords, line, re.IGNORECASE)))
+            if filter(None, found):
+              output = path + '/' + filename + ':' + ','.join(
```
tomchop: Please use a format string here (e.g. `'{0:s}/{1:s}:{2:s}'.format(path, filename, ','.join(filter(None, found))`)

Also, TIL about `filter(None, <list>)`! I would've done `[item for item in <list> if item]`

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+        self.state.add_error(exception, critical=True)
+      # Catch all remaining errors since we want to gracefully report them
+      except Exception as exception:  # pylint: disable=broad-except
+        self.state.add_error(exception, critical=True)
+
+  def grepPDF(self, path):
+    """
+    Parse PDF files text content for keywords.
+
+    :param path: PDF file path.
```
tomchop: dfTimewolf follows the [Google Python style guide](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-method) for docstrings, please adjust

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+    :return: match: set of unique occurrences of every match
+    """
+    with open(path, 'rb') as pdf_file_obj:
+      match = set()
+      text = ""
```
tomchop:
```suggestion
      text = ''
```

Please be consistent with the string quote chars (i.e. pick either single or double quotes and stick to that).

I know this is currently a bit messy throughout the project but let's not make it worse :)

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+              if self._final_output:
+                self._final_output += '\n' + output
+              else:
+                self._final_output = output
+              print(output)
+      except OSError as exception:
+        self.state.add_error(exception, critical=True)
+      # Catch all remaining errors since we want to gracefully report them
+      except Exception as exception:  # pylint: disable=broad-except
+        self.state.add_error(exception, critical=True)
```
tomchop:
```suggestion
        self.state.add_error(exception, critical=True)
        return
```

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+            if filter(None, found):
+              output = path + '/' + filename + ':' + ','.join(
+                  filter(None, found))
+              if self._final_output:
+                self._final_output += '\n' + output
+              else:
+                self._final_output = output
+              print(output)
+      except OSError as exception:
+        self.state.add_error(exception, critical=True)
```
tomchop:
```suggestion
        self.state.add_error(exception, critical=True)
        return
```

dftimewolf/lib/processors/grepper.py
```diff
@@ -0,0 +1,92 @@
...
+    :return: match: set of unique occurrences of every match
+    """
+    with open(path, 'rb') as pdf_file_obj:
+      match = set()
+      text = ""
+      pdf_reader = PyPDF2.PdfFileReader(pdf_file_obj)
+      pages = pdf_reader.numPages
+      for page in range(pages):
+        page_obj = pdf_reader.getPage(page)
+        text += "\n" + page_obj.extractText()
```
tomchop:
```suggestion
        text += '\n' + page_obj.extractText()
```

(btw it's OK to leave the + here since you're only [concatenating two strings](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#310-strings))

dftimewolf/cli/recipes/artifact_grep.py
pstirparo: Done

dftimewolf/cli/recipes/artifact_grep.py
pstirparo: Fixed.
Those are keywords only. Good point to have option for regex as well, I had thought about the file containing keywords. I will add them both for the next PR (I've other functionalities in mind to add as well). Thanks for the suggestions.

dftimewolf/lib/processors/grepper.py
pstirparo: Fixed

dftimewolf/lib/processors/grepper.py
pstirparo: fixed, sorry

dftimewolf/lib/processors/grepper.py
pstirparo: Fixed

dftimewolf/lib/processors/grepper.py
pstirparo: Fixed

dftimewolf/lib/processors/grepper.py
pstirparo: Fixed

dftimewolf/lib/processors/grepper.py
pstirparo: Fixed, thanks

dftimewolf/lib/processors/grepper.py
pstirparo: fixed, both the string the the suggestion. thanks

PR #128, opened 2019-01-09
Turbinia bugfix

tomchop:


tomchop:
@aarontp This is failling because the version that gets installed through pypi is not the one on github HEAD.

PR #132, opened 2019-02-04
Better modules part 2

tomchop:
* Migrate recipes to new module format (add `wants` field)
* Add tests that make sure recipes have the correct syntax


tests/cli/recipes.py
```diff
@@ -0,0 +1,76 @@
...
+        grr_fetch_files,
+        grr_flow_download,
+        grr_hunt_artifacts,
+        grr_hunt_file,
+        grr_huntresults_plaso_timesketch,
+        local_plaso,
+        timesketch_upload,
+    ]
+
+  def testRecipeHasFields(self):
```
Onager: Please add a test for deadlocked modules

PR #136, opened 2019-02-20
Better task printing in turbinia module

tomchop:


dftimewolf/lib/processors/turbinia.py
```diff
@@ -102,26 +153,49 @@ def process(self):
...
+          'instance': self.instance,
+          'project': self.project,
+          'region': self.turbinia_region,
+          'request_id': request.request_id
+      }
+      self.display_task_progress(**request_dict)
+      task_data = self.client.get_task_data(**request_dict)
+      message = 'Completed {0:d} Turbinia tasks\n'.format(len(task_data))
+      for task in task_data:
+        message += '{0:s} ({1:s}): {2:s}\n'.format(
```
tomchop: Should be changed to call `_print_task_data`

PR #137, opened 2019-02-28
Add thread-safe method to state store object

tomchop:
And use them in the Turbinia processor.

### DIFF
--- a/dftimewolf/lib/processors/turbinia.py
+++ b/dftimewolf/lib/processors/turbinia.py
@@ -102,25 +102,50 @@ class TurbiniaProcessor(BaseModule):
     request = TurbiniaRequest()
     request.evidence.append(evidence_)
 
+    # Get threat intelligence data from any modules that have stored some.
+    observables = self.state.get_data('threat_intelligence')
+    if observables:
+      print('Sending {0:d} observables to Turbinia GrepWorkers...'.format(
+          len(observables)))
+      request.recipe['filter_patterns'] = [obs for _, obs in observables]
+
+    request_dict = {
+        'instance': self.instance,
+        'project': self.project,
+        'region': self.turbinia_region,
+        'request_id': request.request_id
+    }
+
     try:
       print('Creating Turbinia request {0:s} with Evidence {1!s}'.format(
           request.request_id, evidence_.name))
       self.client.send_request(request)
       print('Waiting for Turbinia request {0:s} to complete'.format(
           request.request_id))
-      self.client.wait_for_request(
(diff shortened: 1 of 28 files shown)


dftimewolf/lib/processors/turbinia.py
```diff
@@ -102,26 +102,51 @@ def process(self):
     request = TurbiniaRequest()
     request.evidence.append(evidence_)
 
+    # Get threat intelligence data from any modules that have stored some.
+    observables = self.state.get_data('threat_intelligence')
+    if observables:
```
Onager: As discussed, adding a comment here as to what "observables" is supposed to look like would be nice. Perhaps also make an Observable class to make this more checkable

dftimewolf/lib/processors/turbinia.py
```diff
@@ -102,26 +102,51 @@ def process(self):
...
       return
 
+    # Turbinia run complete, build a human-readable message of results.
+    message = 'Completed {0:d} Turbinia tasks\n'.format(len(task_data))
+    for task in task_data:
+      message += '{0!s} ({1!s}): {2!s}\n'.format(
+          task.get('name'),
+          task.get('id'),
+          task.get('status', 'No task status'))
+      for path in task.get('saved_paths') or []:
```
Onager: task.get('save_paths', [])

dftimewolf/lib/processors/turbinia.py
```diff
@@ -102,26 +102,51 @@ def process(self):
...
+        if path.endswith('worker-log.txt'):
+          continue
+        if path.endswith('{0!s}.log'.format(task.get('id'))):
+          continue
+        if path.startswith('/'):
+          continue
+        message += '  {0:s}\n'.format(path)
+
+    # Store the message for consumption by any reporting modules.
+    self.state.store_data('report', message)
```
Onager: What if multiple plugins both want to report something? Change to turbinia_report?

dftimewolf/lib/processors/turbinia.py
tomchop: Done.

dftimewolf/lib/processors/turbinia.py
tomchop: Added a comment to explain that `save_paths` may be populated with `None`.

dftimewolf/lib/processors/turbinia.py
tomchop: As discussed, we'll change this with https://github.com/log2timeline/dftimewolf/issues/138

PR #141, opened 2019-03-05
Beautiful Turbinia task tracking

tomchop:


tomchop:
Ignoring codecov since a big chunk of this code will be deprecated after the change in task display is pushed to Turbinia

PR #144, opened 2019-03-14
Support arg tuples.

someguyiknow:


dftimewolf/lib/utils.py
```diff
@@ -150,6 +150,9 @@ def import_args_from_dict(value, args, config):
         key: import_args_from_dict(val, args, config)
         for key, val in value.items()
     }
+  elif isinstance(value, tuple):
+    return (import_args_from_dict(value[0], args, config),
```
tomchop: This will only work on two-tuples. You can do a tuple comprehension 

```python
return tuple(import_args_from_dict(val, args, config) for val in value)
```

PR #147, opened 2019-03-18
Change VM creation call

tomchop:


tomchop:
This PR is failing because the turbinia pip package with the new function has not been pushed yet.

PR #152, opened 2019-06-04
[DRAFT] Use pipenv to generate requirements.txt

tomchop:


tomchop:
This is not ready for review, but it was the only way to trigger the travis build

Issue #161, opened 2019-06-05
Split module loader from config


joachimmetz:
Split module loader from config so that modules can use the loader to load other modules

joachimmetz:
The split has been completed.

PR #166, opened 2019-06-05
 Added recipe manager #161

joachimmetz:
Added recipe manager #161


joachimmetz:
per conversation https://github.com/log2timeline/dftimewolf/issues/176

PR #169, opened 2019-06-06
 Changes to use JSON recipe files #161

joachimmetz:
Changes to use JSON recipe files #161


joachimmetz:
@tomchop I've updated the PR, I've also cleaned up the dftimewolf tool a bit in this PR. Next steps there would be add better error reporting and tests

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -68,92 +55,182 @@
...
-      recipe['name'], len(recipe['modules'])))
+    configuration_file_path = os.path.join(
+        '/', 'usr', 'share', 'dftimewolf', 'dftimewolf.conf')
+    config.Config.load_extra(configuration_file_path)
 
-  print('Setting up modules...')
-  state.setup_modules(args)
-  print('Modules successfully set up!')
+  def ParseArguments(self, arguments):
+    """Parses the command line arguments.
```
tomchop: This function does more than just parse arguments, namely calls `setup_modules` which already calls logic from each individual module. Maybe rename or extract the `setup_module` call?

joachimmetz: per conversation, I've changed the function.

joachimmetz:
@tomchop PTAL

joachimmetz:
@tomchop PTAL, review status got reset by force push

PR #173, opened 2019-06-07
Added modules manager

joachimmetz:
Added modules manager


tomchop:
Let's merge this and deal with style in #176

PR #175, opened 2019-06-07
Changed collectors to use module manager

joachimmetz:
Changed collectors to use module manager


tomchop:
There's a minor conflict & linter issue, please fix.

PR #180, opened 2019-06-21
Added processor for making Stackdriver logs readable in Timesketch

Onager:


dftimewolf/cli/recipes/stackdriver_gce_instance_timesketch.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""Puts GCE logs into Stackdriver."""
```
tomchop: Is this really what the recipe does?

dftimewolf/lib/processors/stackdriver_timesketch.py
```diff
@@ -0,0 +1,232 @@
...
+    if timestamp:
+      timesketch_record['datetime'] = timestamp
+      timesketch_record['timestamp_desc'] = 'Event Recorded'
+
+    # General resource information.
+    resource = log_record.get('resource', None)
+    if resource:
+      labels = resource.get('labels', None)
+      if labels:
+        for attribute, value in iter(labels.items()):
```
tomchop: is `iter` necessary here?

dftimewolf/lib/processors/stackdriver_timesketch.py
```diff
@@ -0,0 +1,232 @@
...
+  def __init__(self, state):
+    super(StackdriverTimesketch, self).__init__(state)
+
+  def setup(self, *args, **kwargs):
+    """Sets up necessary module configuration options."""
+
+  def cleanup(self):
+    """Cleans up module output to prepare it for the next module."""
+
+  def _process_log_line(self, log_line, query, project_name):
```
tomchop: I wonder if the log parsing logic would not be better implemented in an external module (like `utils` or something similar), so that they can eventually be reused (maybe?) and don't pollute the dftimewolf module class.

tomchop:
Please resolve conflicts and merge or rebase from master to include the latest requirements file

dftimewolf/cli/recipes/stackdriver_gce_instance_timesketch.py
Onager: Not even close - changed this.

dftimewolf/lib/processors/stackdriver_timesketch.py
Onager: Only for Python 2 compatibility, which doesn't matter here.

dftimewolf/lib/processors/stackdriver_timesketch.py
Onager: This is in a separate processor module, so we can extract it somewhere if we need to in the future, but it's pretty specific to Stackdriver->Timesketch mapping.

Issue #176, opened 2019-06-07
have dfTimewolf use log2timeline style guide


joachimmetz:
* [x] change method names to CamelCase
  * https://github.com/log2timeline/dftimewolf/pull/192
  * https://github.com/log2timeline/dftimewolf/pull/193
* [x] update pylintrc to enforce style
  * https://github.com/log2timeline/dftimewolf/pull/194
* [x] update l2tdocs style guide
  * https://github.com/log2timeline/l2tdocs/pull/44
* [x] deprecate timewolf style guide (if present)
  * per tomchop no additional documentation

joachimmetz:
@tomchop is there dfTimewolf specific style guide documentation?

PR #187, opened 2019-07-22
Changes to use log2timeline style guide and removed unused methods #176

joachimmetz:
* remove unused method `_get_client_by_id`
* changes to docstrings
* renamed methods to match log2timeline style guide


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -254,12 +249,11 @@ class GRRArtifactCollector(GRRFlow):
 
   def __init__(self, state):
     super(GRRArtifactCollector, self).__init__(state)
+    self._clients = []
     self.artifacts = []
     self.extra_artifacts = []
     self.hostnames = None
-    self._clients = []
     self.use_tsk = False
-    self.keepalive = False
```
joachimmetz: note for reviewer this is already set by the parent class

PR #190, opened 2019-07-23
Raise exception when no recipe is provided

tomchop:


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -164,6 +164,11 @@ def ParseArguments(self, arguments):
     self._AddRecipeOptions(argument_parser)
 
     self._command_line_options = argument_parser.parse_args(arguments)
+
+    if not hasattr(self._command_line_options, 'recipe'):
```
joachimmetz: what if recipe is set but None? Maybe:

```
if not getattr(self._command_line_options, 'recipe', None):
```

dftimewolf/lib/errors.py
```diff
@@ -11,3 +11,6 @@ def __init__(self, message=None):
 
 class RecipeParseError(DFTimewolfError):
   """Error when parsing a recipe."""
+
```
joachimmetz: `+1` white line

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -221,8 +224,10 @@ def Main():
     print('{0!s}'.format(exception))
     return False
 
-  if not tool.ParseArguments(sys.argv[1:]):
-    # TODO: print errors if this fails.
+  try:
+    tool.ParseArguments(sys.argv[1:])
+  except errors.CommandLineParseError as exception:
+    sys.stderr.write(str(exception))
```
joachimmetz: Isn't sys.stderr.write() expecting bytes (on Python 3)?

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Interesting, I was pretty sure the default return value for `getattr` if no third argument was provided was already None. I should revisit other codebases 😅. Thanks!

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Apparently not, write seems to expect `str`

```
$ python3
Python 3.7.3 (default, Mar 27 2019, 09:23:15)
[Clang 10.0.1 (clang-1001.0.46.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.stderr.write(b'asd')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: write() argument must be str, not bytes
```

tomchop:
Done, PTAL

dftimewolf/cli/dftimewolf_recipes.py
joachimmetz: It likely is, but I prefer it to be specific

tomchop: Ack, thanks for the clarification.

PR #193, opened 2019-07-24
Renamed add_error to AddError to use log2timeline style guide #176

joachimmetz:


dftimewolf/lib/exporters/timesketch.py
```diff
@@ -86,10 +86,10 @@ def Process(self):
...
     try:
       self.timesketch_api.ExportArtifacts(named_timelines, self.sketch_id)
-    except RuntimeError as e:
-      self.state.add_error(
-          'Error occurred while working with Timesketch: {0:s}'.format(str(e)),
-          critical=True)
+    except RuntimeError as exception:
+      self.state.AddError(
+          'An error occurred while working with Timesketch: {0!s}'.format(
+              exception), critical=True)
```
tomchop: Do we have to indent 4 spaces for every subsequent level of indentation?

PR #199, opened 2019-07-25
Add overwrite boolean

tomchop:
This makes it easier to register recipes and modules that have been declared in separate files without knowing their names in advance.

### DIFF
--- a/dftimewolf/lib/recipes/manager.py
+++ b/dftimewolf/lib/recipes/manager.py
@@ -13,9 +13,17 @@ from dftimewolf.lib import resources
 
 
 class RecipesManager(object):
-  """Recipes manager."""
+  """Recipes manager.
+
+  Attribs:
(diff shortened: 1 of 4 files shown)


dftimewolf/lib/recipes/manager.py
```diff
@@ -13,10 +13,18 @@
 
 
 class RecipesManager(object):
-  """Recipes manager."""
+  """Recipes manager.
+
+  Attribs:
+    enable_overwrite: Whether an exception is raised if an existing module is
```
joachimmetz: I opt to change this to a class constant named `ALLOW_RECIPE_OVERRIDE` and please add some tests

tomchop:
Nits fixed, tests added. PTAL

tomchop:
Squash merging since reviewer LGTM'd it

PR #202, opened 2019-07-31
Validate evidence prior to making request

aarontp:
Note that this is dependent on https://github.com/google/turbinia/pull/415 being submitted.


aarontp:
FYI, google/turbinia#415 has been submitted

PR #210, opened 2019-08-27
Add more parameters to Turbinia input

aarontp:
This adds the sketch id to the input parameters for the Turbinia processing module so that we can pass it along.  This will eventually get written out as a metadata file along with the output evidence from Turbinia.  This also excludes the Strings jobs by default (because it takes a long time to run and the output isn't used that often), and adds a --run_all_jobs flag in case we actually want to run the extra jobs.

This also sets the requester information in the TurbiniaRequest.


dftimewolf/lib/processors/turbinia.py
```diff
@@ -50,15 +51,19 @@ def __init__(self, state, critical=False):
...
   # pylint: disable=arguments-differ
-  def SetUp(self, disk_name, project, turbinia_zone):
+  def SetUp(self, disk_name, project, turbinia_zone, sketch_id, run_all_jobs):
     """Sets up the object attributes.
 
     Args:
       disk_name (str): name of the disk to process.
       project (str): name of the GPC project containing the disk to process.
       turbinia_zone (str): GCP zone in which the Turbinia server is running.
+      sketch_id (str): The Timesketch sketch id
```
tomchop: Is sketch_id really a string here? It seems to be ints in the tests.

PR #213, opened 2019-09-04
Add support for HTTP Basic authentication to Timesketch API client

csash:
https://github.com/google/timesketch/pull/403 added the ability to use HTTP Basic authentication to the Timesketch API client in the Timesketch project. This PR updates the Timesketch API client here to do so as well. This PR also updates TimesketchExporter and recipes using it to support the new "auth_mode" parameter on the API client.


dftimewolf/lib/timesketch_utils.py
```diff
@@ -16,44 +16,56 @@ class TimesketchApiClient(object):
     session (requests.Session): HTTP session for calls to Timesketch
   """
 
-  def __init__(self, host_url, username, password):
+  def __init__(self, host_url, username, password, auth_mode=u'timesketch'):
```
berggren: remove the u'

PR #218, opened 2019-09-27
Extract disk name picking to a function

tomchop:
This CL extracts picking of which disk names to copy to a function, so that we can override it if needed.

### DIFF
--- a/dftimewolf/lib/collectors/gcloud.py
+++ b/dftimewolf/lib/collectors/gcloud.py
@@ -159,6 +159,50 @@ class GoogleCloudCollector(module.BaseModule):
           'Try running:\n  $ gcloud auth application-default login')
       self.state.AddError(exception, critical=True)
 
+  def _GetDisksFromNames(self, disk_names):
+    """Gets disks from a project by disk name.
+
+    Args:
+      disk_names (list[str]): List of disk names to get from the project.
+
+    Returns:
+      list[GoogleComputeDisk]: List of GoogleComputeDisk objects to copy.
+    """
+    disks = []
+    for name in disk_names:
+      try:
+        disks.append(self.remote_project.get_disk(name))
+      except RuntimeError:
+        self.state.AddError(
+            'Disk "{0:s}" was not found in project {1:s}'.format(
+                name, self.remote_project.project_id),
+            critical=True)
+    return disks
+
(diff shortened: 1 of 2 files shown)


dftimewolf/lib/collectors/gcloud.py
```diff
@@ -137,43 +154,48 @@ def SetUp(self,
           image_project=image_project,
           image_family=image_family)
 
-      if disk_names:
-        for name in disk_names:
+    except AccessTokenRefreshError as err:
```
Onager: "as exception" per style guide

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -137,43 +154,48 @@ def SetUp(self,
...
+      self.state.AddError('Something is wrong with your gcloud access token.')
+      self.state.AddError(err, critical=True)
+
+    except ApplicationDefaultCredentialsError as err:
+      self.state.AddError(
+          'Something is wrong with your Application Default Credentials. '
+          'Try running:\n  $ gcloud auth application-default login')
+      self.state.AddError(err, critical=True)
+
+  def FindDisksToCopy(self):
```
Onager: Could this be private (_FindDisks...) and return a list of disks? It'd be a bit easier to follow the code, test etc.

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -137,43 +154,48 @@ def SetUp(self,
...
+    """Determines which disks to copy depending on object attributes."""
+    if not (self.remote_instance_name or self.disk_names):
+      self.state.AddError(
+          'You need to specify at least an instance name or disks to copy',
+          critical=True)
+      return
+
+    try:
```
Onager: This is pretty heavily nested and hard to follow, can you please refactor this? Maybe it's worth breaking out some functions here - _GetDisksFromNames() maybe?

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -184,5 +206,8 @@ def SetUp(self,
             'disk name?')
```
Onager: It would be helpful to have the name of the resource you're trying to retrieve here

tomchop: The full detailed exception is now also added in the output.

PR #220, opened 2019-10-09
Added preflight modules

Onager:


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -183,6 +183,11 @@ def ParseArguments(self, arguments):
     print('Loaded recipe {0:s} with {1:d} modules'.format(
         self._recipe['name'], number_of_modules))
 
+  def RunPreflights(self):
+    """Runs preflight modules."""
+    print('Running preflights...')
+    self._state.RunPreflights()
```
tomchop: Consider inlining this (and other functions that follow this pattern) in a future PR.

dftimewolf/lib/module.py
```diff
@@ -42,3 +42,21 @@ def Process(self):
...
+  other modules. They are intended for modules that primarily retrieve
+  attributes for other modules in a recipe, for example from a ticketing system.
+  """
+
+  def SetUp(self, *args, **kwargs):
+    """Preflight modules only have one public method."""
+
+  def Process(self):
+    """Processes input and builds the module's output attribute."""
+
```
tomchop: Maybe re-use these `SetUp` and `Process` functions instead of adding a new one?

dftimewolf/cli/dftimewolf_recipes.py
Onager: Ack, added an issue for this.

dftimewolf/lib/module.py
Onager: Done

PR #231, opened 2019-10-10
Verify TLS connections to TS (fixes #212)

### LINKED ISSUE #212: TLS certificate verification disabled for Timesketch API client
TLS certificate verification is disabled in the Timesketch API client by this line:
https://github.com/log2timeline/dftimewolf/blob/d50f5fb910ad3c50f75151252fb63afc364aafc4/dftimewolf/lib/timesketch_utils.py#L43

It should probably be enabled by default with the ability to disable, like TLS certificate validation for the GRR server URL.

tomchop:


berggren:
Please add a --unsafe flag to let users disable for whatever reason.

PR #238, opened 2019-10-13
Made timesketch endpoint variable name consistent

Onager:


dftimewolf/lib/processors/turbinia.py
```diff
@@ -86,8 +86,8 @@ def SetUp(self, disk_name, project, turbinia_zone, sketch_id, run_all_jobs):
       self.instance = turbinia_config.PUBSUB_TOPIC
       if turbinia_config.TURBINIA_PROJECT != self.project:
         self.state.AddError(
-            'Specified project {0:s} does not match Turbinia configured '
-            'project {1:s}. Use gcp_turbinia_import recipe to copy the disk '
+            'Specified project {0!s} does not match Turbinia configured '
+            'project {1!s}. Use gcp_turbinia_import recipe to copy the disk '
```
tomchop: Any reason we're doing an implicit converstion to `str` here? These should be strings already according to the docstrings.

PR #241, opened 2019-11-12
Add TicketAttribute container

tomchop:
* Add a generic `TicketAttribute` container to be used against ticketing systems
* Have the `GoogleCloudCollector` module store a `TicketAttribute` container with the generated analysis VM information

### DIFF
--- a/dftimewolf/lib/collectors/gcloud.py
+++ b/dftimewolf/lib/collectors/gcloud.py
@@ -10,6 +10,7 @@ from oauth2client.client import ApplicationDefaultCredentialsError
 from turbinia.lib import libcloudforensics
 
 from dftimewolf.lib import module
+from dftimewolf.lib.containers import containers
 from dftimewolf.lib.modules import manager as modules_manager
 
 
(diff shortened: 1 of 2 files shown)


dftimewolf/lib/collectors/gcloud.py
```diff
@@ -31,8 +32,15 @@ class GoogleCloudCollector(module.BaseModule):
         VM will be based.
     all_disks (bool): True if all disks attached to the source
         instance should be copied.
+    ANALYSIS_VM_CONTAINER_ATTRIBUTE_NAME (str): The name of the TicketAttribute
```
Onager: You don't need to add these as attributes - and shouldn't these be private constants anyway, ie. _ANALYSIS_VM_CONTAINER_ATTRIBUTE_NAME ?

tomchop: They can indeed be private since the goal is to subclass this and override these attribute names.

tomchop:
Merging since comments are fixed

PR #243, opened 2019-11-20
Updating gcloud.py to add labels to VM and disks using the incident_id cli argument and adding required unit-tests

sa3eed3ed:
Updating dftimewolf/lib/collectors/gcloud.py to add labels to the VM and all its attached disks and updating tests/lib/collectors/gcloud.py to test for the newly called add_labels function.
This request should be accepted only after https://github.com/google/turbinia/pulls/459 gets accepted, since the libcloudforensics library needs to be updated first.

Unit tests (run_tests.py):
All the existing and the newly added tests passed.
For that I used a local setup referencing turbina.libcloudforensics from https://github.com/sa3eed3ed/turbinia, where the needed function (add_labels) in libcloudforensics is already pushed.

E-2-E testing:
rebuilt dftimewolf after modification and used a recipe that relies on the modified GoogleCloudCollector class to create an analysis_vm in a GCP project and attach the copied disks to it; the created VM and disks get the new incident_id label using the incident_id cli argument as expected.


tomchop:
Added a small modification to the code, but otherwise this looks good. Will merge after the libcloudforensics changes are in and Travis tests stop failing.

tomchop:
Currently failing because of

```
************* Module dftimewolf.lib.collectors.gcloud
dftimewolf/lib/collectors/gcloud.py:66:6: E1101: Instance of 'GoogleComputeDisk' has no 'add_labels' member (no-member)
dftimewolf/lib/collectors/gcloud.py:163:6: E1101: Instance of 'GoogleComputeInstance' has no 'add_labels' member (no-member)
```

Which will be fixed once https://github.com/google/turbinia/pull/459 is in

PR #244, opened 2019-11-21
New grr_artifact_turbinia recipe

aarontp:
This does the following:
- Creates a new recipe `grr_artifact_turbinia`
- Creates new FSPath and RemoteFSPath container types
- Adds `util.Compress()` method to create tarballs
- Changes `LocalFileSystemCopy` module to add a `compress` attribute which will cause it to make a tarball of the output.
- Refactors `TurbiniaProcessor` into `Turbinia{ProcessorBase,GCPProcessor,ArtifactProcessor}` in order to support different types of Turbinia Evidence request types.
- Changes `ScpExporter` to use new container types.


data/recipes/grr_artifact_turbinia.json
```diff
@@ -0,0 +1,75 @@
...
+            "grr_password": "@grr_password",
+            "artifacts": "@artifacts",
+            "extra_artifacts": "@extra_artifacts",
+            "use_tsk": "@use_tsk",
+            "approvers": "@approvers",
+            "verify": "@verify"
+        }
+    }, {
+      "wants": ["GRRArtifactCollector"],
+      "name": "LocalFilesystemCopy",
```
tomchop: I *think* you can omit the LocalFilesystemCopy module here, unless you explicitly want the files to be copied to in some user-defined directory.

aarontp: This is still required in this case because the local filesystem module is what I'm using to do the compression.

aarontp:
BTW, the test error doesn't seem to be related to my changes AFAICT.  Seems like more dependency conflicts? :(

dftimewolf/lib/exporters/local_filesystem.py
```diff
@@ -44,8 +50,18 @@ def SetUp(self, target_directory=None):  # pylint: disable=arguments-differ
...
-      print('{0:s} -> {1:s}'.format(path, self._target_directory))
+      if not self._compress:
+        full_paths = self._CopyFileOrDirectory(path, self._target_directory)
+        print('{0:s} -> {1:s}'.format(path, self._target_directory))
+        for path_ in full_paths:
+          self.state.StoreContainer(containers.FSPath(path=path_))
+      else:
+        try:
+          tar_file = utils.Compress(path, self._target_directory)
+          self.state.StoreContainer(containers.FSPath(path=tar_file))
```
tomchop: For consistency, could you add a print statement like above to indicate where the files are going?

dftimewolf/lib/exporters/scp_ex.py
```diff
@@ -47,28 +49,47 @@ def SetUp(self, # pylint: disable=arguments-differ
...
+    cmd = ['scp']
     cmd.extend(self._paths)
     cmd.append(dest)
     print('Executing SCP command: {0:s}'.format(' '.join(cmd)))
     ret = subprocess.call(cmd)
     if ret != 0:
-      self.state.AddError("Failed copying {0!s}".format(self._paths),
+      self.state.AddError('Failed copying {0!s}'.format(self._paths),
                           critical=True)
+    else:
```
tomchop: It would be cool to change the check on `ret` to be a guard statement (replace this `else` by `return` and unindent the code below?)

dftimewolf/lib/processors/turbinia_artifact.py
```diff
@@ -0,0 +1,73 @@
...
+from turbinia import TurbiniaException
+
+from dftimewolf.lib.containers import containers
+from dftimewolf.lib.modules import manager as modules_manager
+from dftimewolf.lib.processors.turbinia_gcp import TurbiniaProcessorBase
+
+# pylint: disable=no-member
+
+class TurbiniaArtifactProcessor(TurbiniaProcessorBase):
+  """Processes Google Cloud (GCP) disks with Turbinia.
```
tomchop: This processes local artifacts / files, rather than GCP disks, correct?

dftimewolf/lib/processors/turbinia_artifact.py
```diff
@@ -0,0 +1,73 @@
...
+    self.directory_path = None
+
+  # pylint: disable=arguments-differ
+  def SetUp(
+      self, directory_path, project, turbinia_zone, sketch_id, run_all_jobs):
+    """Sets up the object attributes.
+
+    Args:
+      directory_path (str): Name of the directory to process.
+      project (str): Name of the GCP project containing the disk to process.
```
tomchop: Isn't this the project where Turbinia lives?

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -43,79 +44,69 @@ def __init__(self, state, critical=False):
...
-    self.run_all_jobs = None
 
   # pylint: disable=arguments-differ
-  def SetUp(self, disk_name, project, turbinia_zone, sketch_id, run_all_jobs):
+  def TurbiniaSetUp(self, project, turbinia_zone, sketch_id, run_all_jobs):
     """Sets up the object attributes.
 
     Args:
-      disk_name (str): name of the disk to process.
       project (str): name of the GPC project containing the disk to process.
```
tomchop: This is the project where Turbinia lives? Admittedly, it contains the disks to process, but it would be clearer IMO to explicitly assume that this is the case.

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -43,79 +44,69 @@ def __init__(self, state, critical=False):
...
-      # elsewhere in the codebase.
-      self.state.AddError(exception, critical=True)
+    self.turbinia_region = turbinia_config.TURBINIA_REGION
+    self.instance = turbinia_config.PUBSUB_TOPIC
+    if turbinia_config.TURBINIA_PROJECT != self.project:
+      self.state.AddError(
+          'Specified project {0!s} does not match Turbinia configured '
+          'project {1!s}. Use gcp_turbinia_disk_copy_ts recipe to copy the '
+          'disk into the same project.'.format(
+              self.project, turbinia_config.TURBINIA_PROJECT), critical=True)
```
tomchop: This check becomes moot if we specify that the project is actually the project where turbinia lives. UX-wise, I don't think it makes sense here to let the user pick the project they wanna use (this can be tweaked in the recipes that invoke this module)

dftimewolf/lib/exporters/local_filesystem.py
aarontp: Done.

dftimewolf/lib/exporters/scp_ex.py
aarontp: Done.

dftimewolf/lib/processors/turbinia_artifact.py
aarontp: Yes!  Thanks.  Fixed.

dftimewolf/lib/processors/turbinia_artifact.py
aarontp: Yeah.  This can actually be refactored out for this module.  I removed the attributes for now, but I'll take a second pass to refactor this a bit in this PR before submitting it.

dftimewolf/lib/processors/turbinia_gcp.py
aarontp: Yeah, I'll fix this up a bit, thanks.

dftimewolf/lib/processors/turbinia_gcp.py
aarontp: Yeah, this has actually changed a little bit on the Turbinia side.   By default it should just read the project/zone from the config, and if the values that are passed in to Turbinia are different from the config, it will assume that the disk needs to be copied into the correct project/zone and will copy it over.  This works now because it happens client-side and just uses the same libcloudforensics method that dftimewolf uses.  I'll clean this up a bit on the dftimewolf side to make it more obvious what is going and remove the unneeded params.

dftimewolf/lib/processors/turbinia_artifact.py
```diff
@@ -0,0 +1,98 @@
...
+      for task in task_data:
+        saved_paths = task.get('saved_paths') or []
+        for path in saved_paths:
+          # Ignore temporary files generated by turbinia
+          if path.startswith('/tmp'):
+            continue
+
+          # We're only interested in plaso files for the time being.
+          if path.endswith('.plaso'):
+            path = path.replace('/var/lib/turbinia', '/mnt/turbinia/output')
```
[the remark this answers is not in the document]
aarontp: We could set that as an attribute from the base class Turbinia setup method.

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -149,26 +113,59 @@ def _DownloadFilesFromGCS(self, timeline_label, gs_paths):
...
-    vm_containers = self.state.GetContainers(containers.ForensicsVM)
-    if vm_containers and not self.disk_name:
-      forensics_vm = vm_containers[0]
-      self.disk_name = forensics_vm.evidence_disk.name
-      self.logger.info(
-          'Using disk {0:s} from previous collector'.format(self.disk_name))
+  def TurbiniaSetUp(self, project, turbinia_zone, sketch_id, run_all_jobs):
+    """Sets up the object attributes.
+    Args:
+      project (str): name of the GPC project containing the disk to process.
```
aarontp, line 126: s/GPC/GCP/.   Also, super minor nit, but you probably want a newline above the args section.

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -149,26 +113,59 @@ def _DownloadFilesFromGCS(self, timeline_label, gs_paths):
...
+      self.turbinia_zone = turbinia_config.TURBINIA_ZONE
+
+    if self.project is None or self.turbinia_zone is None:
+      self.ModuleError(
+          'project or turbinia_zone are not all specified, bailing out',
+          critical=True)
+      return
+
+    self.turbinia_region = turbinia_config.TURBINIA_REGION
+    self.instance = turbinia_config.PUBSUB_TOPIC
```
aarontp, line 149: I think we want `turbinia_config.INSTANCE_ID` here now.  It used to be the same as pubsub topic, and it still is often set to that, but it's a separate variable that can sometimes be different.

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -227,6 +225,72 @@ def Process(self):
...
+            project,
+            turbinia_zone,
+            sketch_id,
+            run_all_jobs):
+    """Sets up the object attributes.
+
+    Args:
+      turbinia_config_file (str): Full path to the Turbinia config file to use.
+      disk_name (str): name of the disk to process.
+      project (str): name of the GPC project containing the disk to process.
```
aarontp, line 262: s/GPC/GCP/

tests/lib/processors/turbinia_gcp.py
```diff
@@ -81,23 +83,27 @@ def testWrongProject(self, _mock_TurbiniaClient):
...
 
+  @mock.patch('dftimewolf.lib.processors.turbinia_gcp.turbinia_config')
   @mock.patch('turbinia.client.TurbiniaClient')
-  def testWrongSetup(self, _mock_TurbiniaClient): # pylint: disable=invalid-name
+  # pylint: disable=invalid-name
+  def testWrongSetup(self, _mock_TurbiniaClient, mock_turbinia_config):
     """Tests that invalid setup options generate errors."""
     params = [
         {
+             'turbinia_config_file': None,
```
aarontp, line 98: nit: whitespace (here and below)

aarontp (commented):
Other than some minor things (most of which was probably my fault in the first place :)), it LGTM!

(I would hit "approve" but it won't let me since I'm the original author)

data/recipes/upload_turbinia.json
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }, {
+    "wants": ["FilesystemCollector"],
+    "name": "LocalFilesystemCopy",
+    "args": {
+        "target_directory": "@directory",
+        "compress": true
+    }
+  }, {
+        "wants": ["LocalFilesystemCopy"],
```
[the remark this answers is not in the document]
tomchop: I'll have a JSON formatting pass on recipes in another PR, most of them are all wonky.

dftimewolf/lib/processors/turbinia_artifact.py
```diff
@@ -0,0 +1,98 @@
...
+
+class TurbiniaArtifactProcessor(TurbiniaProcessorBase):
+  """Processes Exported GRR Artifacts with Turbinia.
+
+  Attributes:
+    directory_path (str): Name of the directory to process.
+  """
+
+  def __init__(self, state, name=None, critical=False):
+    """Initializes a Turbinia Google Cloud (GCP) disks processor.
```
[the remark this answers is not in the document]
tomchop: Done :)

dftimewolf/lib/processors/turbinia_artifact.py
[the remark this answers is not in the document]
tomchop: Yes, this was a temporary workaround until Turbinia provided the right paths in its output. I'll remove this logic for now and add the recipe attribute in another PR if we still need it.

tomchop:
I messed up this PR. Fixed reviewer comments in https://github.com/log2timeline/dftimewolf/pull/350/commits/538448aa438cbabc6cd862a4e0f141d2a9ec1a24, PTAL there. Closing this one.

PR #245, opened 2019-11-21
Fix access to None object attribute when flow cannot be launched

tomchop:


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -328,6 +332,11 @@ def _ProcessThread(self, client):
         ignore_interpolation_errors=True,
         apply_parsers=False)
     flow_id = self._LaunchFlow(client, 'ArtifactCollectorFlow', flow_args)
+    if not flow_id:
+      msg = 'Flow could not be launched on {0:s}.'.format(client.client_id)
+      msg += '\nArtifactCollectorFlow args: {0:s}'.format(flow_args)
```
aarontp: Do you need this format string to be '{0!s}' since flow_args isn't actually a string?

tomchop: Correct!

PR #248, opened 2020-01-21
Introduce StreamingContainers

tomchop:
This PR introduces StreamingContainers in dfTimewolf.

Modules register StreamingContainersConsumers by passing a callback and an AttributeContainer class to `RegisterStreamingCallback`. Other modules then call `StreamContainer` with the container of their choice, and all callbacks registered for that container type will be called. 

Containers "streamed" this way are not stored and cannot be recovered later on, unless the callback implements an internal mechanism to preserve them.

### DIFF
--- a/dftimewolf/lib/state.py
+++ b/dftimewolf/lib/state.py
@@ -221,7 +221,7 @@ class DFTimewolfState(object):
       self.streaming_callbacks[container.CONTAINER_TYPE] = []
     self.streaming_callbacks[container.CONTAINER_TYPE].append(target)
 
-  def StoreStreamingContainer(self, container):
+  def StreamContainer(self, container):
     """Streams a container to the callbacks that are registered to handle it.
 
     Args:
--- a/dftimewolf/lib/state.py
+++ b/dftimewolf/lib/state.py
@@ -227,7 +227,6 @@ class DFTimewolfState(object):
     Args:
       container (interface.AttributeContainer): container that will be streamed
           to any registered callbacks.
-
     """
     for container_type, callbacks in self.streaming_callbacks.items():
       if container_type == container.CONTAINER_TYPE:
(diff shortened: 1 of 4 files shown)


dftimewolf/lib/state.py
```diff
@@ -208,6 +209,30 @@ def RunModules(self):
     """Performs the actual processing for each module in the module pool."""
     self._InvokeModulesInThreads(self._RunModuleThread)
 
+  def RegisterStreamingCallback(self, target, container):
+    """Registers a callback for a type of container.
+
+    Args:
+      target (function): function to be called.
```
Onager: Please add the method signature you're expecting here.

dftimewolf/lib/state.py
```diff
@@ -208,6 +209,30 @@ def RunModules(self):
...
+      container (interface.AttributeContainer): container type on which
```
Onager: It's actually the container type that's relevant here, so "container whose type..."

But maybe this method should just take the type instead?

dftimewolf/lib/state.py
```diff
@@ -208,6 +209,30 @@ def RunModules(self):
...
+    self.streaming_callbacks[container.CONTAINER_TYPE].append(target)
+
+  def StreamContainer(self, container):
+    """Streams a container to the callbacks that are registered to handle it.
+
+    Args:
+      container (interface.AttributeContainer): container that will be streamed
+          to any registered callbacks.
+    """
+    for container_type, callbacks in self.streaming_callbacks.items():
```
Onager: callbacks = self.streaming_callbacks.get([container.CONTAINER_TYPE], [])

tomchop: Wow, yeah, I don't know what I was thinking.

dftimewolf/lib/state.py
tomchop: We're technically passing a type (e.g. `container.Report`). It just happens that AttributeContainers have an attribute that describes their type as strings. I changed this to use the actual types instead of the attribute (see f5f8f14)

tomchop:
Changes had been LGTM'd so I'm merging this (f5f8f14 being the only new change)

PR #257, opened 2020-03-18
Add e2e test for libcloudforensics.

giovannt0:

### DIFF
--- a/tests/lib/collectors/gcloud_e2e.py
+++ b/tests/lib/collectors/gcloud_e2e.py
@@ -20,8 +20,8 @@ import unittest
 import logging
 
 from googleapiclient.errors import HttpError
-
 from libcloudforensics import gcp
+
 from dftimewolf import config
 from dftimewolf.lib import state
 from dftimewolf.lib.collectors import gcloud
@@ -57,45 +57,24 @@ class EndToEndTest(unittest.TestCase):
 
   def __init__(self, *args, **kwargs):
     super(EndToEndTest, self).__init__(*args, **kwargs)
-    project_info = os.environ.get('PROJECT_INFO')
-
-    if project_info is None:
-      self.error_msg = 'Error: please make sure that you defined the ' \
-                       '"PROJECT_INFO" environment variable pointing ' \
-                       'to your project settings.'
-      return
     try:
-      json_file = open(project_info)
-      project_info = json.load(json_file)
-      json_file.close()
-    except ValueError as exception:
-      self.error_msg = 'Error: cannot parse JSON file. {0:s}'.format(
-          str(exception))
+      project_info = ReadProjectInfo()
+    except (OSError, RuntimeError, ValueError) as exception:
+      self.error_msg = str(exception)
       return
-
-    if not all(key in project_info for key in ['project_id', 'instance',
-                                               'zone']):
-      self.error_msg = 'Error: please make sure that your JSON file ' \
-                       'has the required entries. The file should ' \
-                       'contain at least the following: ["project_id", ' \
-                       '"instance", "zone"].'
-      return
-
     self.project_id = project_info['project_id']
     self.instance_to_analyse = project_info['instance']
     # Optional: test a disk other than the boot disk
-    if 'disk' in project_info:
-      self.disk_to_forensic = project_info['disk']
-    else:
-      self.disk_to_forensic = None
+    self.disk_to_forensic = project_info.get('disk', None)
     self.zone = project_info['zone']
+    self.gcp = gcp.GoogleCloudProject(self.project_id, self.zone)
 
   def setUp(self):
     if hasattr(self, 'error_msg'):
       raise unittest.SkipTest(self.error_msg)
-    self.analysis_vm = None
-    self.test_state = None
     self.incident_id = 'fake-incident-id'
+    self.test_state = state.DFTimewolfState(config.Config)
+    self.gcloud_collector = gcloud.GoogleCloudCollector(self.test_state)
 
   def test_end_to_end_boot_disk(self):
     """End to end test on GCP for the gcloud.py collector.
@@ -109,12 +88,8 @@ class EndToEndTest(unittest.TestCase):
       analysis purpose and attach the boot disk copy to it.
     """
 
-    self.test_state = state.DFTimewolfState(config.Config)
-    gcloud_collector = gcloud.GoogleCloudCollector(self.test_state)
-    self.assertIsNotNone(gcloud_collector)
-
-    # This should make a copy of the boot disk only
-    gcloud_collector.SetUp(
+    # Setting up the collector to make a copy of the boot disk only
+    self.gcloud_collector.SetUp(
         self.project_id,
         self.project_id,
         self.incident_id,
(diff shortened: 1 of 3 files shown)


tests/lib/collectors/libcloudforensics_e2e.py
```diff
@@ -0,0 +1,246 @@
...
+
+    disks = self.analysis_vm.list_disks()
+    self.assertEqual(disks,
+                     ['gcp-forensics-vm-fake-incident-id',
+                      self.test_state.output[0][1].name,
+                      self.test_state.output[1][1].name])
+
+    self.__clean()
+
+  def __clean(self):
```
tomchop: Consider the case where we'll have more than one end-to-end test (e.g with different parameters passed in SetUp()), you may want to call this method `tearDown()` ([link](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown)) as it will be called after each test method invocation (the same happens with `setUp()`)

giovannt0: Addressed, I separated the test in two for better readability.

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+          str(exception))
+      return
+
+    if not all(key in project_info for key in ['project_id', 'instance',
+                                               'zone']):
+      self.error_msg = 'Error: please make sure that your JSON file ' \
+                       'has the required entries. The file should ' \
+                       'contain at least the following: ["project_id", ' \
+                       '"instance", "zone"].'
+      return
```
tomchop: Please extract this to its an independent function that attempts to parse the JSON file and raises exceptions if it fails, returns correct JSON if it succeeds.

(this function is already quite complex and almost requires testing in itself)

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+
+    self.project_id = project_info['project_id']
+    self.instance_to_analyse = project_info['instance']
+    # Optional: test a disk other than the boot disk
+    if 'disk' in project_info:
+      self.disk_to_forensic = project_info['disk']
+    else:
+      self.disk_to_forensic = None
```
tomchop: You can replace these 4 lines by

```python
self.disk_to_forensic = project_info.get('disk', None)
```

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+    self.zone = project_info['zone']
+
+  def setUp(self):
+    if hasattr(self, 'error_msg'):
+      raise unittest.SkipTest(self.error_msg)
```
tomchop: Will `SkipTest` actually generate an error in the test run? Or will it say something like "yay tests succeeded, 3 / 3 tests skipepd"

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+      1. The gcloud.py collector connects to the target instance and makes a
+      snapshot of the boot disk.
+      2. A new disk is created from the taken snapshot.
+      3. If an analysis VM already exists, the module will attach the disk
+      copy to the VM. Otherwise, it will create a new GCP instance for
+      analysis purpose and attach the boot disk copy to it.
+    """
+
+    self.test_state = state.DFTimewolfState(config.Config)
+    gcloud_collector = gcloud.GoogleCloudCollector(self.test_state)
```
tomchop: This can probably co in SetUp(), as you're repeating the same lines for each test

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+    self.assertIsNotNone(gcloud_collector)
```
tomchop: No need to test this here as we're already testing it in unittests

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+    # This should make a copy of the boot disk only
+    gcloud_collector.SetUp(
+        self.project_id,
+        self.project_id,
+        self.incident_id,
+        self.zone,
+        42.0,
+        16,
+        remote_instance_name=self.instance_to_analyse,
+    )
```
tomchop: I would add a comment saying we expect the parameter `disk_names` to default to `None` here

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+
+  def __assertCollectorIsInitialized(self, gcloud_collector):
+    self.assertEqual(self.test_state.errors, [])
+    self.assertEqual(gcloud_collector.analysis_project.project_id,
+                     self.project_id)
+    self.assertEqual(gcloud_collector.remote_project.project_id,
+                     self.project_id)
+    self.assertEqual(gcloud_collector.remote_instance_name,
+                     self.instance_to_analyse)
+    self.assertEqual(gcloud_collector.all_disks, False)
```
tomchop: We probably don't need this as this should all be tested in the unittests

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -0,0 +1,253 @@
...
+    self.assertIsInstance(self.test_state.output[0][1], gcp.GoogleComputeDisk)
+    self.assertTrue(self.test_state.output[0][1].name.startswith(
+        'incidentfake-incident-id'))
+    self.assertTrue(self.test_state.output[0][1].name.endswith('-copy'))
+    self.assertIsInstance(self.analysis_vm, gcp.GoogleComputeInstance)
+    self.assertEqual(self.analysis_vm.name, 'gcp-forensics-vm-fake-incident-id')
+    self.assertEqual(
+        self.analysis_vm.list_disks(),
+        ['gcp-forensics-vm-fake-incident-id',
+         self.test_state.output[0][1].name])
```
tomchop: I think here we should be comparing dftimewolf's internal state with the things that are actually happening in GCP.

* Check that the VM that dftimewolf asked to create actually exists in the gcp project (by calling the external API)
* Check that the disks that are being created are the ones we expect (also by calling the external API)

giovannt0:
Thanks for the comments, addressed in the latest commit.

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -109,28 +88,42 @@ def test_end_to_end_boot_disk(self):
...
-
     # Attach the boot disk copy to the analysis VM
-    gcloud_collector.Process()
-    self.__assertInstanceAndDisk()
+    self.gcloud_collector.Process()
+
+    # The forensic instance should be live in the analysis GCP project and
+    # the disk should be attached
+    analysis_vm_name = self.test_state.output[0][0]
+    attached_disk_name = self.test_state.output[0][1].name
```
tomchop: Maybe call this variable `expected_disk_name`?

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -109,28 +88,42 @@ def test_end_to_end_boot_disk(self):
...
+      result = self.gcp.gce_operation(operation, zone=self.zone)
+
+      self.assertEqual(result['name'], analysis_vm_name)
+      for disk in result['disks']:
+        if disk['source'].split("/")[-1] == attached_disk_name:
+          return
+      self.fail('Error: could not find the disk {0:s} in instance {1:s}'.format(
+          attached_disk_name, analysis_vm_name
+      ))
+    except RuntimeError as exception:
```
tomchop: What in the code is expected to raise a `RuntimeError`? Make sure  to narrow down the try / except blocks as much as possible. Also, it might be fine to just not try to catch this exception as we'll get a full stacktrace on where the error came from (easier for debugging, which is what tests are for)

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -161,51 +150,42 @@ def test_end_to_end_other_disk(self):
...
+      self.fail('Error: could not find the forensic instance in the analysis '
+                'project: {0:s}'.format(str(exception)))
```
tomchop: Same comment about the try / except statement as above

tests/lib/collectors/gcloud_e2e.py
```diff
@@ -161,51 +150,42 @@ def test_end_to_end_other_disk(self):
...
-    disks = self.analysis_vm.list_disks()
+  def tearDown(self):
+    project = self.gcloud_collector.analysis_project
+    disks = self.gcloud_collector.analysis_vm.list_disks()
 
     # delete the created forensics VMs
     log.info('Deleting analysis instance: {0:s}.'.format(
-        self.analysis_vm.name))
+        self.gcloud_collector.analysis_vm.name))
     operation = project.gce_api().instances().delete(
```
tomchop: For consistency it's better that you use the `self.gcp` attribute you defined in L70

tests/lib/collectors/gcloud_e2e.py
giovannt0: The calls to the GCP API done by the `gcp` module would raise a RuntimeError if the API call fails, but I agree this can be discarded.

PR #261, opened 2020-03-31
Add public CleanUp method in gcloud_e2e test.

giovannt0:
This allows e2e test from the CLI and gives the possibility to clean up the created VMs afterwards by calling the module's `CleanUp` method directly.

### DIFF
--- a/tests/lib/collectors/gcloud_e2e.py
+++ b/tests/lib/collectors/gcloud_e2e.py
@@ -170,60 +170,11 @@ class EndToEndTest(unittest.TestCase):
     ))
 
   def tearDown(self):
-    disks = self.gcloud_collector.analysis_vm.ListDisks()
-
-    # delete the created forensics VMs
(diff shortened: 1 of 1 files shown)


tests/lib/collectors/gcloud_e2e.py
```diff
@@ -170,60 +170,12 @@ def test_end_to_end_other_disk(self):
...
-          if exception.resp.status != 400:
-            log.warning('Could not delete the disk {0:s}: {1:s}'.format(
-                disk, str(exception)
-            ))
-          # Throttle the requests to one every 10 seconds
-          time.sleep(10)
-
-      log.info('Disk {0:s} successfully deleted.'.format(
-          disk))
+    CleanUp(project_id=self.project_id, zone=self.zone,
```
tomchop: These are positional arguments, and don't require the keword argument notation:

```python
CleanUp(self.project_id, self.zone, self.gcloud_collector.analysis_vm.name)
```

PR #266, opened 2020-05-04
Support All FileFinder Actions

someguyiknow:


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -406,6 +413,9 @@ def SetUp(self,
     self.hostnames = [item.strip() for item in hosts.strip().split(',')]
     self.use_tsk = use_tsk
 
+    if action.lower() in self._ACTIONS:
+      self.action = self._ACTIONS[action.lower()]
+
```
tomchop, line 418: Would probably be better to bail if the action specified is not in `self._ACTIONS`

```python
self.action = self._ACTIONS.get(action.lower())
if not self.action:
  # add critical error
```

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -416,12 +426,12 @@ def _ProcessThread(self, client):
       client (object): GRR client object to act on.
     """
     file_list = self.files
-    if not file_list:
+    if not file_list or self.action is None:
```
tomchop, line 429: remove the check on self.action if you're bailing during SetUp

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -395,6 +401,7 @@ def SetUp(self,
       approvers (Optional[str]): list of GRR approval recipients.
       verify (Optional[bool]): True to indicate GRR server's x509 certificate
           should be verified.
+      action (Optional[str]): Action to take (download/hash/stat).
```
tomchop, line 404: + "(default: download)"

dftimewolf/lib/collectors/grr_hosts.py
someguyiknow: Done. Had to use is None because stat enum is 0.

tomchop:
Thanks!

PR #267, opened 2020-05-05
Renamed stackdriver to gcp_logging, removed use of oauthclient

Onager:


dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -85,16 +83,15 @@ def Process(self):
...
       self.state.AddError(
           'Something is wrong with your gcloud access token.')
       # TODO: determine if exception should be converted into a string as
       # elsewhere in the codebase.
       self.state.AddError(exception, critical=True)
       return
 
-    except (ApplicationDefaultCredentialsError,
-            google_auth_exceptions.DefaultCredentialsError) as exception:
+    except google_auth_exceptions.DefaultCredentialsError as exception:
```
tomchop, line 94: Could we also catch `google_auth_exceptions.RefreshError` here? Both errors are probably solved by refreshing Application Default Credentials.

Onager: Sure, done.

PR #274, opened 2020-05-13
Changing Timesketch behavior: using the API client and importer client.

kiddinn:
This PR changes the behavior in dftimewolf to start using the Timesketch API and importer clients instead of implementing their own API client.

This changes also makes use of latest changes in the API client to store configuration settings to save user input.


kiddinn:
tests run locally on my machine

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -22,86 +27,130 @@ class TimesketchExporter(module.BaseModule):
...
+    if not sketch:
+      sketch = self.timesketch_api.get_sketch(self.sketch_id)
 
-    sketch_url = self.timesketch_api.GetSketchUrl(self.sketch_id)
+    recipe_name = self.state.recipe.get('name', 'no_recipe')
+    input_names = []
+    for description, _ in self.state.input:
+      if not description:
+        continue
+      name, _, _ = description.rpartition('.')
```
tomchop, line 117: isn't `name = description.rpartition('.')[0]` a bit clearer?

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -22,86 +27,130 @@ class TimesketchExporter(module.BaseModule):
...
+      name = name.replace(' ', '_')
+      name = name.replace('-', '_')
```
tomchop, line 119: You can chain these

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -22,86 +27,130 @@ class TimesketchExporter(module.BaseModule):
...
+
+    with importer.ImportStreamer() as streamer:
+      streamer.set_sketch(sketch)
+      streamer.set_timeline_name(timeline_name)
+
+      for _, path in self.state.input:
+        streamer.add_file(path)
+
+    api_root = sketch.api.api_root
+    host_url, _, _ = api_root.partition('api/v1')
```
tomchop, line 136: Same, can you use `[0]` here?

dftimewolf/lib/exporters/timesketch.py
kiddinn: I'm not sure I agree with it being clearer, but nonetheless no strong feelings here so done....

dftimewolf/lib/exporters/timesketch.py
kiddinn: yes... just a question which one is more readable... chained

dftimewolf/lib/exporters/timesketch.py
kiddinn: done

kiddinn:
done with this round, PTAL @tomchop

PR #277, opened 2020-05-18
Adding new  gce_disk_export recipe

sa3eed3ed:
Adding new  gce_disk_export recipe to export GCE disks to GCS buckets.
Adding new GoogleCloudDiskExport module.
Adding unittest for the new recipe.
Fixing few doc-strings.


data/recipes/gce_disk-export.json
```diff
@@ -0,0 +1,24 @@
...
+        "args": {
+            "analysis_project_name": "@analysis_project_name",
+            "source_project_name": "@source_project_name",
+            "source_disk_name": "@source_disk_name",
+            "gcs_output_location": "@gcs_output_location",
+            "exported_disk_name": "@exported_disk_name"
+        }
+    }],
+    "args": [
+        ["--analysis_project_name", "Project where the disk image is created then exported. If not exist, source_project_name will be used.", null],
```
tomchop, line 17: please put optional args at the end. I don't think it makes any difference in the code, but it's easier to understand what parameters the recipe needs.

dftimewolf/lib/collectors/gce_disk_export.py
```diff
@@ -0,0 +1,103 @@
...
+    image_object = self.analysis_project.CreateImageFromDisk(
+        self.source_disk, name=self._image_name)
+    image_object.ExportImage(
+        self.gcs_output_location, output_name=self.exported_disk_name)
+    image_object.Delete()
+    output_uri = os.path.join(
+        self.gcs_output_location, '{0:s}.tar.gz'.format(
+            self.exported_disk_name))
+    print('Disk was exported to: {0:s}'.format(output_uri))
+    self.state.output.append((output_uri))
```
tomchop, line 59:
```suggestion
    self.state.output.append(output_uri)
```

dftimewolf/lib/collectors/gce_disk_export.py
```diff
@@ -0,0 +1,103 @@
...
+    """Sets up a Google Cloud Platform (GCP) Disk Export.
+
+    This method creates the required objects to initialize
+    the GoogleCloudDiskExport class attributes.
+
+    If the analysis_project_name is not specified it will use the
+    source_project_name instead.
+
+    Args:
+      analysis_project_name (Optional[str]): Project ID where the
```
tomchop, line 77: This is not really true - `SetUp` will fail if no argument is provided for analysis_project_name (it's not a keyword argument). Like in the recipe, please put all optional parameters at the end and if they're optional, please add a default value in python (regardless of whether a default value is setup in the recipe parameters).

dftimewolf/lib/collectors/gce_disk_export.py
```diff
@@ -0,0 +1,103 @@
...
+        self.gcs_output_location, output_name=self.exported_disk_name)
+    image_object.Delete()
+    output_uri = os.path.join(
+        self.gcs_output_location, '{0:s}.tar.gz'.format(
+            self.exported_disk_name))
+    print('Disk was exported to: {0:s}'.format(output_uri))
+    self.state.output.append((output_uri))
+
+  # pylint: disable=arguments-differ
+  def SetUp(self,
```
tomchop, line 62: I don't know the answer to this, but it would be great if we could ensure that all credentials are good to go during SetUp (that the bucket actually exists, and that we actually have write access to it) , and error out early if they aren't.

Otherwise, if we end up chaining this module after a long-running processing task and have it fail last minute, it won't make for a great user experience.

dftimewolf/lib/collectors/gce_disk_export.py
sa3eed3ed: There is a way to test IAM permissions via projects.testIamPermissions under the resource-manager APIs. This will need direct calls to Cloud APIs and I was thinking to add these to libcloudforensics, before the time consuming  operations. On the other side, the export function will trigger a cloud build job which will fail right away if it doesn't have access to the output bucket, it doesn't wait till the export is done to check. So I think it will save us little time here only if we check the permissions on the output bucket before the createImage function starts to create an image, what do you think?

dftimewolf/lib/collectors/gce_disk_export.py
sa3eed3ed: There is a way to test IAM permissions via projects.testIamPermissions under the resource-manager APIs. This will need direct calls to Cloud APIs and I was thinking to add these to libcloudforensics, before the time consuming operations. On the other side, the export function will trigger a cloud build job which will fail right away if it doesn't have access to the output bucket, it doesn't wait till the export is done to check. So I think it will save us little time here only if we check the permissions on the output bucket before the createImage function starts to create an image, what do you think?

tomchop: I see. The thing is when a dftimewolf recipe is run all the module's `SetUp` functions are called simultaneously, and if one of them fails, then the recipe stops executing. If they succeed, then modules are called in the sequence defined by the recipe. In this case if this module comes after other time-consuming modules, and it fails because of some IAM permissions, then the user will only find out at the very end. 

I'm fine to wait for changes in LCF before implementing this check, but it's still something to think about.

dftimewolf/lib/collectors/gce_disk_export.py
```diff
@@ -0,0 +1,103 @@
...
+from __future__ import unicode_literals
+
+import os
+from libcloudforensics import gcp
+
+from dftimewolf.lib import module
+from dftimewolf.lib.modules import manager as modules_manager
+
+
+class GoogleCloudDiskExport(module.BaseModule):
```
tomchop, line 14: Also with a name like this, it should probably go in `/exporters/`

sa3eed3ed: Yes, that makes sense, will move it to exporters.

dftimewolf/lib/collectors/gce_disk_export.py
sa3eed3ed: yes it would be definitely better to check in advance, I did couple of trials but there are few things that makes this tricky. The cloud build function runs using the cloud build service account in the analysis project, we need to figure out the project number to be able to construct the service account name, moreover to check the permissions for that service account, we need to run as this service account which will require a different permission than the ones needed for that recipe, there might be a workaround using the access-troubleshooter APIs and making the permission check in the SetUp function optional in case the user have enough permissions to access the troubleshooter service but this idea is still experimental. I would keep it like that for now and open a new issue to add these checks. How does that sound to you?

data/recipes/gce_disk-export.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "gce_disk_export",
```
tomchop, line 2: Please make sure the filename matches the recipe name

PR #287, opened 2020-06-10
Fix gcloud_e2e.py by upgrading libcloudforensics

giovannt0:


giovannt0:
My pipenv is weird, dropping this PR

Issue #276, opened 2020-05-15
timesketch_api_client error while save credential


knsankar:
Hi,
I pulled latest update (_4159e3bb97e22c1b630f3d717a12f4887693534d_) and tried to run _plaso_ts_, _grr_artifact_ts_ and it was asking for credentials to create timesketchrc. After provided the uri, user, password, I am getting following error.

Configuration loaded from: /opt/dftimewolf/data/config.json
Loading recipe...
Loaded recipe plaso_ts with 3 modules
Running preflights...
Setting up modules...
[WARNING] No timesketch section in the config
What is the value for [auth_mode]: timesketch
What is the value for [host_uri]: http://127.0.0.1:5000
What is the value for [username]: admin
Password for user admin:
[INFO] File does not exist, creating it.
[INFO] Credentials saved to: /root/.timesketch.token
dfTimewolf encountered one or more errors:
CRITICAL:   An unknown error occurred: 'NoneType' object has no attribute 'serialize'
Full traceback:
Traceback (most recent call last):
  File "/opt/dftimewolf/dftimewolf/lib/state.py", line 160, in _SetupModuleThread
    module.SetUp(**new_args)
  File "/opt/dftimewolf/dftimewolf/lib/exporters/timesketch.py", line 54, in SetUp
    self.timesketch_api = timesketch_utils.GetApiClient(self.state)
  File "/opt/dftimewolf/dftimewolf/lib/timesketch_utils.py", line 44, in GetApiClient
    ts_client.credentials, config_assistant=assistant)
  File "/opt/dftimewolf/venv/lib64/python3.6/site-packages/timesketch_api_client/crypto.py", line 257, in save_credentials
    data = cred_obj.serialize()
AttributeError: 'NoneType' object has no attribute 'serialize'

Critical error found. Aborting.

\
\
\

If I ignore this error and I try to run the recipe again, i am getting a different, wrong credential error.
\
\
\

Configuration loaded from: /opt/dftimewolf/data/config.json
Loading recipe...
Loaded recipe plaso_ts with 3 modules
Running preflights...
Setting up modules...
dfTimewolf encountered one or more errors:
CRITICAL:   An unknown error occurred: Unable to decrypt data, password wrong? (error )
Full traceback:
Traceback (most recent call last):
  File "/opt/dftimewolf/venv/lib64/python3.6/site-packages/cryptography/fernet.py", line 104, in _verify_signature
    h.verify(data[-32:])
[...]

PR #296, opened 2020-07-06
Changing the TS importer behavior

kiddinn:
Changing the behavior of the TS importer:

1. Fix an issue with the importer, #276 
2. Add the option to supply custom token passwords
3. Add more error reporting


kiddinn:
Requires https://github.com/google/timesketch/pull/1288 to be pushed and uploaded to pypi

data/recipes/gcp_logging_cloudaudit_ts.json
```diff
@@ -18,6 +18,7 @@
         "name": "TimesketchExporter",
         "args": {
             "incident_id": "@incident_id",
+	    "token_password": "@token_password",
```
tomchop, line 21: Plz fix indentation here and in the other json files.

data/recipes/gcp_logging_gce_ts.json
```diff
@@ -26,5 +27,6 @@
         ["start_date", "Start date (yyyy-mm-ddTHH:MM:SSZ)", null],
         ["end_date", "End date (yyyy-mm-ddTHH:MM:SSZ)", null],
         ["--incident_id", "Incident ID (used for Timesketch description)", null],
+        ["--token_password", "Optional custom password to decrypt Timesketch credential file with", ""],
```
tomchop, line 30: Just for consistency, do you mind putting this one after `sketch_id`?

dftimewolf/lib/timesketch_utils.py
```diff
@@ -32,14 +34,23 @@ def GetApiClient(state):
...
-    ts_client = assistant.get_client()
+    config.configure_missing_parameters(
+        config_assistant=assistant, token_password=token_password)
+
+    ts_client = assistant.get_client(token_password=token_password)
+
+    if not ts_client:
+      state.AddError(
+          'Unable to get a Timesketch API Client', critical=False)
+      return ts_client
```
tomchop, line 45: return `None` would probably be better here.

dftimewolf/lib/timesketch_utils.py
```diff
@@ -9,7 +9,7 @@
 LOCK = threading.Lock()
 
 
-def GetApiClient(state):
+def GetApiClient(state, token_password=''):
```
tomchop, line 12: Shouldn't this default to `None`? It feels like encrypting with an empty password should not be the same as not encrypting the file. (unless you're XORing 😅)

data/recipes/gcp_logging_cloudaudit_ts.json
kiddinn: hmm... the indentation looks "correct" in my editor... let's see

it seems like a `<TAB>` got somehow introduced there in between spaces... fixed

dftimewolf/lib/timesketch_utils.py
kiddinn: done

dftimewolf/lib/timesketch_utils.py
kiddinn: well, no, the password is a string, and the upstream library expects a string to be passed in, and if the string is empty then the `default` pwd is used, which is auto generated

tomchop: Ack, sounds good. Could you add a bit of docstring that says this?

kiddinn: sure can....

PR #300, opened 2020-07-08
Pull extra attributes from GCP Logging

Fryyyyy:
The policy deltas in particular


Fryyyyy:
Testing fails with:
```Warning: Python 3.7 was not found on your system…
Neither 'pyenv' nor 'asdf' could be found to install Python.
You can specify specific versions of Python with:
$ pipenv --python path/to/python```
???

tomchop:
Tests are failing because pipenv is unhappy with the python version that's running in 3.8 (we're specifying 3.7 in the pipfile). I'll fix this in another PR, since `pip` tests are passing.

PR #303, opened 2020-07-09
Changing TS exporter to reflect API changes

kiddinn:
Recent changes in TS changed the return object from analyzer run.


tomchop:
Ignoring pipenv failures as this is a known issue. pip and travis are passing.

PR #304, opened 2020-07-09
Better logging and color display in terminals

tomchop:
This PR brings some color into the very monotone world of wolves.

Two main additions:

A `SetupLogging` function in `dftimewolf_recipes.py`, to initialize the main dftimewolf logging handlers.
A `logging_utils.py` module where the color sequences are generated and the log formatter is defined.

The rest of the changes are mainly moves from `print` to `self.logger.info()` (with the occasional `error`)


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -37,6 +38,11 @@
 
 from dftimewolf.lib.recipes import manager as recipes_manager
 from dftimewolf.lib.state import DFTimewolfState
+from dftimewolf.lib import logging_utils
+
+logger = logging.getLogger('dftimewolf')
+DEFAULT_LOG_FILE = os.path.join('/', 'tmp', 'dftimewolf.log')
```
kiddinn, line 44: what about using `os.sep` instead of `/` here?

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -134,11 +141,11 @@ def _LoadConfigurationFromFile(self, configuration_file_path):
     """
     try:
       if config.Config.LoadExtra(configuration_file_path):
-        sys.stderr.write('Configuration loaded from: {0:s}\n'.format(
+        logger.debug('Configuration loaded from: {0:s}'.format(
             configuration_file_path))
 
     except errors.BadConfigurationError as exception:
-      sys.stderr.write('{0!s}'.format(exception))
+      logger.warning('{0!s}'.format(exception))
```
kiddinn, line 148: is this a `warning` or an `error` ?

also if you want to include the stack trace you can also just simply do:

```
except errors.BadConfigurationError:
  logger.warning('Bad configuration value, with error:', exc_info=True)
```

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -225,27 +233,57 @@ def SignalHandler(*unused_argvs):
   sys.exit(0)
 
 
+def SetupLogging():
+  """Sets up a logging handler with dftimewolf's custom formatter."""
+  # Clear root handlers (Turbinia is setting them)
```
kiddinn, line 238: maybe you can just say "dependent libraries are configuring the root logger', since turbina may change it's way and other libraries may be messing with it too....

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -225,27 +233,57 @@ def SignalHandler(*unused_argvs):
...
+  root_log = logging.getLogger()
+  root_log.handlers = []
+
+  # Add and silence a default stream handler, this is automatically set
```
kiddinn, line 242: s/Add and silence a default stream handler/Add a silence default stream handler/

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -225,27 +233,57 @@ def SignalHandler(*unused_argvs):
...
+  # when other libraries call logging.info() or similar methods.
+  root_handler = logging.StreamHandler()
+  root_handler.addFilter(lambda x: False)
+  root_log.addHandler(root_handler)
+
+  # We want all error messages. Maybe make this customizable in the future?
```
kiddinn, line 248: technically you are not gettingn all `error` messages, but DEBUG and above... you're getting all messages. and maybe add a TODO here to have this configurable.

There could be a default value of INFO and a `--verbose`, `--debub` parameter to the tool tha twoul dchange it from INFO level to DEBUG

dftimewolf/lib/logging_utils.py
```diff
@@ -0,0 +1,71 @@
+"""Module providing custom logging formatters."""
+import logging
+import random
+
+def _GenerateColorSequences():
```
kiddinn, line 5: doesn't this only work for Bash? or at least for Linux terminals? Can we make that somewhere clear?

What about if you run th etool on a non-Linux box?

dftimewolf/lib/logging_utils.py
```diff
@@ -0,0 +1,71 @@
...
+  """Generates ANSI codes for 256 colors."""
+  sequences = []
+  for i in range(0, 16):
```
kiddinn, line 8: `for i ...` and `for j ....` this feels like a copy/paste from stack overflow... can we at least name these variables (style guide)

dftimewolf/lib/logging_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Cherrypick a few interesting values. We still want the whole list of colors
+# so that modules have a good amount colors to chose from.
+# pylint: disable=unbalanced-tuple-unpacking
+BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE = COLOR_SEQS[8:16]
+BG_RED = '\u001b[41m'  # Red background
+BOLD = '\u001b[1m'  # Bold / bright modifier
+
+LOG_FORMAT = ('[%(asctime)s] [{0:s}{color:s}%(name)-20s{1:s}] %(levelname)-8s'
+              ' %(message)s')
```
kiddinn, line 26: what about `%(name)s/%(levelname)s` instead of just `%levelname`....

that is the `name` gives you the name of the logger. In the case of timesketch api/importer, etc thsi would give you where this is coming from (since you already `killed` the root logger, you should only get back records from a named logger)

dftimewolf/lib/state.py
```diff
@@ -12,6 +13,7 @@
 from dftimewolf.lib import utils
 from dftimewolf.lib.modules import manager as modules_manager
 
+logger = logging.getLogger('dftimewolf')
```
kiddinn, line 16: I would arge to use here:

```
...getLogger('dftimewolf.state')
```

and that way you can have a more granularity into where the logs are coming from

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Sure, even though I'm pretty sure this will fail on an OS where the `sep` is not `/`.

dftimewolf/cli/dftimewolf_recipes.py
tomchop: In this case it's a warning, since we'd like to continue with defaults. Good tip about the stack trace! Do you know if it's part of the same message or is it split by lines?

dftimewolf/cli/dftimewolf_recipes.py
tomchop: True. Done.

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Done.

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Done.

dftimewolf/lib/logging_utils.py
tomchop: It works on macOS. I haven't tried Windows, but pretty sure WSL supports it. I'll add a note.

dftimewolf/lib/logging_utils.py
tomchop: I honestly don't know what to call them other than counters. This is how the numbers match to the colors: https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit

dftimewolf/lib/logging_utils.py
tomchop: we already have `(name)s` in brackets. This is what it looks like:

```
[2020-07-09 18:06:05,187] [TimesketchExporter  ] INFO     Sketch 23 created
```

`TimesketchExporter` being the name of the logger (named after the class name of the module that initializes it; this can be changed)

dftimewolf/lib/state.py
tomchop: Gotcha. In this case, I don't think it makes much difference for the user to know that this is coming from the state (I'm OK with conflating the `state` with dftimewolf's main execution). WDYT?

dftimewolf/cli/dftimewolf_recipes.py
kiddinn: it's split by lines (unfortunately)

dftimewolf/lib/logging_utils.py
kiddinn: ack

dftimewolf/lib/logging_utils.py
kiddinn: some how I missed the name there... sorry about that ;)

dftimewolf/lib/logging_utils.py
kiddinn: well, that's a good question ... let's leave it as is....

dftimewolf/lib/state.py
kiddinn: it's also for debugging purposes, that is for you to know that the error came up in wha tmodule/part.... your call though

dftimewolf/cli/dftimewolf_recipes.py
tomchop: That's good to know, thanks. In my case I don't want to log the stack trace at the same time as the error, so I store it somewhere for later use.

dftimewolf/lib/state.py
tomchop: Let's leave it as-is for now, if this is something that I need to change later on I will.

tomchop:
Resolved both comments. Added a TODO to consider whether to change the logger name.

PR #312, opened 2020-07-16
☠️

tomchop:


dftimewolf/lib/state.py
```diff
@@ -340,6 +343,11 @@ def CheckErrors(self, is_global=False):
       if error.critical:
         critical_errors = True
 
+    if any(error.unexpected for error in error_objects):
+      logger.critical('One or more unexpected errors occurred.')
```
kiddinn, line 347: don't want to include the `exc_info` in this here? to make the error report more complete?

tomchop: `exc_info` is already printed just before. Printing it twice (almost contiguously) might be too verbose and confusing.

kiddinn: ahh... ok, didn't see that

PR #315, opened 2020-07-21
Better better logging

tomchop:
Add a RotatingFileHandler to the modules logger.

### DIFF
--- a/dftimewolf/cli/dftimewolf_recipes.py
+++ b/dftimewolf/cli/dftimewolf_recipes.py
@@ -4,6 +4,8 @@
 
 import argparse
 import logging
+# Some AttributeErrors occured when trying to access logging.handlers, so
+# we import them separately
 from logging import handlers
 import os
 import signal
--- a/dftimewolf/cli/dftimewolf_recipes.py
+++ b/dftimewolf/cli/dftimewolf_recipes.py
@@ -43,7 +43,6 @@ from dftimewolf.lib.state import DFTimewolfState
 from dftimewolf.lib import logging_utils
 
 logger = logging.getLogger('dftimewolf')
-DEFAULT_LOG_FILE = os.path.join(os.sep, 'tmp', 'dftimewolf.log')
 
 
 class DFTimewolfTool(object):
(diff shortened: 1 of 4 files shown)


dftimewolf/lib/module.py
```diff
@@ -3,6 +3,7 @@
 
 import abc
 import logging
+from logging import handlers
```
Onager, line 6: Why import this again? Just do logging.handlers.

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -254,15 +253,16 @@ def SetupLogging():
   # File handler needs go be added first because it doesn't format messages
   # with color
   file_handler = handlers.RotatingFileHandler(
-      DEFAULT_LOG_FILE, maxBytes=5*1024*1024, backupCount=3)
+      logging_utils.DEFAULT_LOG_FILE, maxBytes=5*1024*1024, backupCount=3)
```
Onager, line 256: Move the maxbytes and backupCount values here to logging_utils as well

dftimewolf/lib/module.py
```diff
@@ -30,10 +31,21 @@ def __init__(self, state, critical=False):
...
     self.critical = critical
     self.state = state
+    self.SetupLogging()
+
+  def SetupLogging(self):
+    """Sets up stream and file logging for a specific module."""
     self.logger = logging.getLogger(name=self.__class__.__name__)
+
+    file_handler = handlers.RotatingFileHandler(
+        logging_utils.DEFAULT_LOG_FILE, maxBytes=5*1024*1024, backupCount=3)
```
Onager, line 41: Move the maxbytes and backupCount values here to logging_utils as well

dftimewolf/lib/module.py
tomchop: I'd love to, but doing so caused some actual runtime errors: https://github.com/log2timeline/dftimewolf/pull/306

Onager: Hmm, that's weird - please add a comment explaining that

PR #317, opened 2020-07-30
PR1/2 Export-> GCS -> Import -> Turbinia -> TS

sa3eed3ed:
This is PR 1 of 2 PRs to add a recipe that export disks from an evidence project GCP to GCS, import them into a different GCP project, attach disks to an analysis VM, run Turbinia and upload results to Timesketch. The following changes are pushed to change the gce_disk_export recipe and allow it to export multiple disks or all the disks attached to a VM at once instead of a single disk only, in addition to that, there are few enhancements and corrections in the current GoogleCloudCollector module and dependent recipes
Changes:
- Disk export recipe to accept a list of disks or an evidence VM as input and export all of them to GCS
- Changing unit-tests
- Create analysis VM is made optional.
- cpu_cores as an optional argument with default of 4
- Recipes description nits
- Incident ID is optional: not affecting the current default behaviour
- if no incident ID default is appending timestamp to the
analysis VM prefix
- Zone is optional for analysis VM


data/recipes/gcp_forensics.json
```diff
@@ -11,22 +11,25 @@
...
-          "cpu_cores": 16,
+          "cpu_cores": "@cpu_cores",
           "image_project": "ubuntu-os-cloud",
           "image_family": "ubuntu-1804-lts"
       }
   }],
   "args": [
       ["remote_project_name", "Name of the project containing the instance / disks to copy ", null],
-      ["incident_id", "Incident ID to label the VM with.", null],
+      ["analysis_project_name", "Name of the project where the analysis VM will be created", null],
```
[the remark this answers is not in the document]
tomchop: I think like you may have wanted to harmonize this and the `gcp_turbinia_disk_copy_ts.json` recipe. It would maybe be better to revisit this in a future PR than to squeeze it in here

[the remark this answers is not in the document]
sa3eed3ed: Got it. I will make the analysis_project_name optional again.

data/recipes/gcp_forensics.json
```diff
@@ -11,22 +11,25 @@
...
+      ["--incident_id", "Incident ID to label the VM with.", null],
```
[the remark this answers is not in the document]
sa3eed3ed: For the "incident-id" I made it optional since I am making the creation of the analysis VM optional, this was a ToDo comment that I ran through in the code and I think it make the recipe more flexible. It will not affect the existing way of invoking this recipe, it will still work. What do you think?

dftimewolf/lib/exporters/gce_disk_export.py
```diff
@@ -72,32 +81,64 @@ def SetUp(self,
...
+    gcloud_collector = gcloud.GoogleCloudCollector(init_state)
+    gcloud_collector.SetUp(
+        analysis_project_name=self.analysis_project.project_id,
+        remote_project_name=self.source_project.project_id,
+        create_analysis_vm=False,
+        disk_names=disk_names,
+        all_disks=all_disks,
+        remote_instance_name=remote_instance_name)
+
+    self.source_disks = gcloud_collector.FindDisksToCopy()
```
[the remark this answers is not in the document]
sa3eed3ed: I agree that it is better to avoid doing things this way, I just wanted to avoid copying FindDisksToCopy since I will need to copy _GetDisksFromInstance and _GetDisksFromNames as well, but I will copy them and do it in a more proper way.

data/recipes/gcp_forensics.json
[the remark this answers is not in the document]
sa3eed3ed: Oh also I remember something else,  gcloud doesn't check if analysis_project_name exist or not, if not it will default to none and then error when it try to print the analysis vm name. I was buggy, I will ad additional condition to make it work as aoptional.

data/recipes/gcp_forensics.json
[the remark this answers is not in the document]
tomchop: I think in the interest of the user, I'd try to make the interface as consistent as possible across the different `gcp_` recipes.

Another point - you only create the `_gcp_label` attribute if an analysis VM is being created, but this label is also applied to disks. In the current logic it will only be applied to disks if the VM is created, which seems to be two different things. It might be better to adjust the code to apply a label based on `incident_id` no matter what

data/recipes/gcp_forensics.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "gcp_forensics",
   "short_description": "Copies disk from a GCP project to an analysis VM.",
-  "description": "Copies a disk from a project to another, creates an analysis VM, and attaches the copied disk to it.",
+  "description": "Copies a disk from a project to another, it will creates an analysis VM, and attaches the copied disk to it.",
```
tomchop, line 4: Why change this? All other sentences are in the present tense.

data/recipes/gcp_forensics.json
```diff
@@ -11,22 +11,25 @@
...
   "args": [
       ["remote_project_name", "Name of the project containing the instance / disks to copy ", null],
-      ["incident_id", "Incident ID to label the VM with.", null],
+      ["--analysis_project_name", "Name of the project where the analysis VM will be created", null],
+      ["--incident_id", "Incident ID to label the VM with.", null],
       ["--instance", "Name of the instance to analyze.", null],
       ["--disks", "Comma-separated list of disks to copy.", null],
       ["--all_disks", "Copy all disks in the designated instance. Overrides disk_names if specified", false],
-      ["--analysis_project_name", "Name of the project where the analysis VM will be created", null],
+      ["--analysis_vm", "Create an analysis VM.", true],
```
tomchop, line 31: I'd rename this to `create_analysis_vm` for more clarity in the UI

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -77,21 +78,25 @@ def Process(self):
...
-            zone,
-            boot_disk_size,
-            boot_disk_type,
-            cpu_cores,
+            incident_id=None,
+            zone='us-central1-f',
+            create_analysis_vm=True,
+            boot_disk_size=50,
+            boot_disk_type='pd-standard',
+            cpu_cores=4,
```
tomchop, line 86: It's better to leave these arguments as positional args, so that they error out if the recipe is not providing everything it needs (a recipe developer will catch errors earlier on).

(in another PR) It might be good to change this across the whole codebase.

dftimewolf/lib/exporters/gce_disk_export.py
```diff
@@ -36,34 +42,42 @@ def __init__(self, state, critical=False):
...
+      self.state.StoreContainer(container)
 
   # pylint: disable=arguments-differ
   def SetUp(self,
             source_project_name,
-            source_disk_name,
             gcs_output_location,
             analysis_project_name=None,
-            exported_disk_name=None):
+            disk_names=None,
```
tomchop, line 77: rename this to `source_disk_names`, so as to be consistent with the previous parameter name.

dftimewolf/lib/exporters/gce_disk_export.py
```diff
@@ -72,32 +86,140 @@ def SetUp(self,
...
     Args:
       source_project_name (str): Source project ID containing
           the disk to export.
-      source_disk_name (str): Disk that needs to be exported.
       gcs_output_location (str): Google Cloud Storage parent bucket/folder
           path of the exported image.
-      analysis_project_name (Optional[str]): Project ID where the
-          disk image is created then exported. If not exit,
+      analysis_project_name (Optional[str]): Optional. Project ID where the
+          disk image is created then exported. Default,
```
tomchop, line 104:
```suggestion
          disk image is created then exported. If not specified,
```

dftimewolf/lib/exporters/gce_disk_export.py
```diff
@@ -72,32 +86,140 @@ def SetUp(self,
...
     self.source_project = gcp_project.GoogleCloudProject(source_project_name)
     if analysis_project_name:
       self.analysis_project = gcp_project.GoogleCloudProject(
           analysis_project_name)
     else:
       self.analysis_project = self.source_project
-    self.source_disk = self.source_project.compute.GetDisk(source_disk_name)
+
+    if not (remote_instance_name or disk_names):
+      self.state.AddError(
```
tomchop, line 129: We've added a `self.ModuleError` function that you should call instead of `AddError`. It takes exactly the same parameters so refactoring should be easy. It will also raise an error, so no need to `return` after critical errors.

tests/lib/exporters/gce_disk_export.py
```diff
@@ -39,35 +39,42 @@ def testInitialization(self):
 
   # pylint: disable=line-too-long
   @mock.patch('libcloudforensics.providers.gcp.internal.compute.GoogleCloudCompute.GetDisk')
+  @mock.patch('dftimewolf.lib.collectors.gcloud.GoogleCloudCollector.SetUp')
```
tomchop, line 42: Why are we mocking GoogleCloudCollector if we're testing the exporter here?

tomchop (changes requested):

data/recipes/gcp_forensics.json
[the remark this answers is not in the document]
sa3eed3ed: both GCP recipes now have the --incident_id attribute as optional, so now its consistent. 
The disks will also get _gcp_label added however this will happen in the process method, while the created VM will have the _gcp_label in the startup method, but both are covered

data/recipes/gcp_forensics.json
sa3eed3ed: changed it back

data/recipes/gcp_forensics.json
sa3eed3ed: Done

dftimewolf/lib/collectors/gcloud.py
sa3eed3ed: will open an issue for this and do it in a separate PR in all places

dftimewolf/lib/exporters/gce_disk_export.py
sa3eed3ed: Done

dftimewolf/lib/exporters/gce_disk_export.py
sa3eed3ed: Done

dftimewolf/lib/exporters/gce_disk_export.py
sa3eed3ed: Done

tests/lib/exporters/gce_disk_export.py
sa3eed3ed: removed, was a residue from before, where I used a dummy gcloud collector to find instance disks, but now no need.

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -59,11 +59,13 @@ def Process(self):
       new_disk = gcp_forensics.CreateDiskCopy(
           self.remote_project.project_id,
           self.analysis_project.project_id,
+          None,
```
tomchop, line 62: This might be some weird merge artifact, but i removed the extraneous `None` which has disappeared from the latest version of LCF: https://github.com/google/cloud-forensics-utils/blob/e6e53f8bbf2eec86e6e4f4a54bcddf6ead7902d3/libcloudforensics/providers/gcp/forensics.py#L34

tomchop: Same thing in the tests for this function

tests/lib/exporters/gce_disk_export.py
```diff
@@ -87,30 +91,26 @@ def testProcess(self,
...
     cloud_disk_exporter.SetUp(
-        'fake-source-project',
-        'fake-source-disk',
-        'gs://fake-bucket',
-        None,
-        None
+        source_project_name='fake-source-project',
+        gcs_output_location='gs://fake-bucket',
+        source_disk_names='fake-source-disk',
+        exported_image_name='image-df-export-temp'\
```
tomchop, line 100: Remove the extra `\`

dftimewolf/lib/exporters/gce_disk_export.py
```diff
@@ -72,32 +86,140 @@ def SetUp(self,
...
+
+    try:
+
+      if self.source_disk_names:
+        disks_to_copy = self._GetDisksFromNames(self.source_disk_names)
+
+      elif self.remote_instance_name:
+        disks_to_copy = self._GetDisksFromInstance(self.remote_instance_name,
+                                                   self.all_disks)
+
```
tomchop, line 206: Please remove the whitespace here

dftimewolf/lib/exporters/gce_disk_export.py
```diff
@@ -72,32 +86,140 @@ def SetUp(self,
...
-    self.source_disk = self.source_project.compute.GetDisk(source_disk_name)
+
+    if not (remote_instance_name or source_disk_names):
+      self.ModuleError(
+          'You need to specify at least an instance name or disks to copy',
+          critical=True)
+
+    self.remote_instance_name = remote_instance_name
+    source_disk_names = source_disk_names.split(
+        ',') if source_disk_names else []
```
tomchop, line 135: The newline here makes it a bit weird to read. Could you keep the `split()` call on the same line?

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -130,56 +144,65 @@ def SetUp(self,
...
+              type_=self._ANALYSIS_VM_CONTAINER_ATTRIBUTE_TYPE,
+              value=analysis_vm_name))
 
-    self.logger.info('Your analysis VM will be: {0:s}'.format(analysis_vm_name))
-    self.logger.info('Complimentary gcloud command:')
-    self.logger.info(
-        'gcloud compute ssh --project {0:s} {1:s} --zone {2:s}'.format(
+      try:
+        # pylint: disable=too-many-function-args
+        # pylint: disable=redundant-keyword-arg
```
tomchop, line 187:
```suggestion
        # pylint: disable=redundant-keyword-arg,too-many-function-args
```

data/recipes/gcp_forensics.json
```diff
@@ -11,22 +11,25 @@
...
       ["remote_project_name", "Name of the project containing the instance / disks to copy ", null],
-      ["incident_id", "Incident ID to label the VM with.", null],
+      ["--analysis_project_name", "Name of the project where the analysis VM will be created", null],
+      ["--incident_id", "Incident ID to label the VM with.", null],
       ["--instance", "Name of the instance to analyze.", null],
       ["--disks", "Comma-separated list of disks to copy.", null],
       ["--all_disks", "Copy all disks in the designated instance. Overrides disk_names if specified", false],
-      ["--analysis_project_name", "Name of the project where the analysis VM will be created", null],
+      ["--create_analysis_vm", "Create an analysis VM.", true],
+      ["--cpu_cores", "Number of CPU cores of the analysis VM", 4],
```
tomchop, line 32: Someone specifying this parameter through the CLI will turn it into an `str`. Please make sure this is converted to an int during `SetUp`

tomchop: Actually, this should be changed in dftimewolf_recipes.py:83. I openeed https://github.com/log2timeline/dftimewolf/issues/341 for that. (but in the meantime, please make sure it's converted to an int)

tomchop (changes requested):

dftimewolf/lib/exporters/gce_disk_export.py
sa3eed3ed: Done

tests/lib/exporters/gce_disk_export.py
sa3eed3ed: done

dftimewolf/lib/exporters/gce_disk_export.py
sa3eed3ed: done

dftimewolf/lib/collectors/gcloud.py
sa3eed3ed: Done

data/recipes/gcp_forensics.json
sa3eed3ed: done in setup method

tomchop:
Sorry to keep you waiting @sa3eed3ed but I didn't see you had pushed changes! I'll have a look at this this week if time permits.

tomchop (approved):
LGTM. Let's wait for e2e tests to complete before merging.

PR #325, opened 2020-08-28
Enable searching for GRR hosts by client ID

tomchop:
* Enable querying hosts by client_id
* change some module-internal variable and function names

TODO: Change the variable names in recipes too.

### DIFF
--- a/dftimewolf/lib/collectors/grr_hosts.py
+++ b/dftimewolf/lib/collectors/grr_hosts.py
@@ -109,7 +109,7 @@ class GRRFlow(GRRBaseModule):  # pylint: disable=abstract-method
     clients = []
     for selector in selectors:
       client = self._GetClientBySelector(selector)
-      if client:
+      if client is not None:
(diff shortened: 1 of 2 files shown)


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -512,7 +516,9 @@ def Process(self):
     Raises:
       DFTimewolfError: if no files specified
     """
-    client = self._GetClientByHostname(self.host)
+    # TODO (tomchop): The host attribute into something more appropriate
```
Onager, line 519: Did you mean "change the host attribute"?

Onager (approved):

dftimewolf/lib/collectors/grr_hosts.py
tomchop: Yes, thanks. I'll push the fix and then merge the PR.

PR #334, opened 2020-09-30
feat: Skip waiting for offline hosts

tomchop:
Add a `skip_offline_hosts` parameter to the GRR recipes so that offline hosts to not block the recipe flow. Offline hosts will be skipped and output will be generated like below, 

```
[2020-09-30 16:35:31,695] [GRRFlowCollector    ] INFO     Searching for client: C.dd8b402e40f260f4
[2020-09-30 16:35:34,178] [GRRFlowCollector    ] INFO     C.dd8b402e40f260f4: Found active client
[2020-09-30 16:35:34,179] [GRRFlowCollector    ] INFO     Found active client: C.dd8b402e40f260f4
[2020-09-30 16:35:34,179] [GRRFlowCollector    ] INFO     Client last seen: 2020-09-30T14:35:07+0000 (0 minutes ago)
[2020-09-30 16:35:34,180] [GRRFlowCollector    ] INFO     ED19A889: Waiting to finish
[2020-09-30 16:35:34,180] [GRRFlowCollector    ] INFO     Client will be skipped if offline.
[2020-09-30 16:35:44,508] [GRRFlowCollector    ] WARNING  Client C.dd8b402e40f260f4 has been offline for more than an hour, skipping...
[2020-09-30 16:35:44,508] [GRRFlowCollector    ] WARNING  Skipped waiting for 1 flows because hosts were offline
[2020-09-30 16:35:44,508] [GRRFlowCollector    ] WARNING  Run the grr_flow_collect recipe with --wait_for_offline_hosts in a tmux shell
[2020-09-30 16:35:44,508] [GRRFlowCollector    ] WARNING  dftimewolf grr_flow_collect C.dd8b402e40f260f4 ED19A889 random /tmp/directory
```


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -176,7 +204,38 @@ def _AwaitFlow(self, client, flow_id):
...
         self.logger.info('{0:s}: Complete'.format(flow_id))
         break
+
       time.sleep(self._CHECK_FLOW_INTERVAL_SEC)
+      if not self.skip_offline_clients:
+        continue
+
+      client_last_seen = datetime.datetime.utcfromtimestamp(
+          client.data.last_seen_at / 1000000)
+      now = datetime.datetime.utcnow()
```
kiddinn, line 214: see: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow 

```
Because naive datetime objects are treated by many datetime
 methods as local times, it is preferred to use aware datetimes
 to represent times in UTC. As such, the recommended way to
 create an object representing the current time in UTC is by
 calling datetime.now(timezone.utc).
```

and: https://docs.python.org/3/library/datetime.html#datetime.datetime.now

```
This function is preferred over today() and utcnow().

```

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -176,7 +204,38 @@ def _AwaitFlow(self, client, flow_id):
...
+      if (now - client_last_seen).total_seconds() > self._MAX_OFFLINE_TIME_SEC:
+        self.logger.warning(
+              'Client {0:s} has been offline for more than an'
+              ' hour, skipping...'.format(
```
kiddinn, line 218: well, that's not necessarily true... it's been more than MAX_OFFLINE_TIME_SEC since the host was still last seen....

so .... if I change the offline time sec variable, or we later change to be configurable, then this will not hold... maybe have this rather as `for more than {self._MAX_OFFLINE_TIME_SEC} seconds` or calculate the minutes `f'for more than {self._MAX_OFFLINE_TIME_SEC / 60} minutes'` or something like that

kiddinn (commented):
few comments

kiddinn:
there is also some pylint issues (as in one)

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -551,7 +621,7 @@ def __init__(self, state):
   def SetUp(self,
             hosts, root_path,
             reason, timeline_format, grr_server_url, grr_username, grr_password,
-            approvers=None, verify=True):
+            approvers=None, verify=True, skip_offline_clients=False):
```
kiddinn, line 624: linter complains:
```

dftimewolf/lib/collectors/grr_hosts.py:621:2: R0913: Too many arguments (11/10) (too-many-arguments)
```

tomchop: Yeah, unfortunately such is the way of using recipes.

dftimewolf/lib/collectors/grr_hosts.py
tomchop: Good point.

dftimewolf/lib/collectors/grr_hosts.py
tomchop: Fixed to use `fromtimestamp()` and `now()` with `datetime.utc` TZ (although since we're only manipulating the dates here it doesn't really matter that much I think)

kiddinn (approved):
LGTM

PR #339, opened 2020-10-02
Move sketch creation logic to Process

tomchop:
* Moves the sketch creation logic to `Process`
* Only check for ACLs in `SetUp`


dftimewolf/lib/exporters/timesketch.py
```diff
@@ -68,28 +68,46 @@ def SetUp(self,  # pylint: disable=arguments-differ
...
+    if incident_id:
+      sketch_name = 'Sketch for incident ID: ' + incident_id
+    else:
+      sketch_name = 'Untitled sketch'
+    sketch_description = 'Sketch generated by dfTimewolf'
+
+    sketch = self.timesketch_api.create_sketch(
+        sketch_name, sketch_description)
+    self.sketch_id = sketch.id
+    self.logger.info('Sketch {0:d} created'.format(self.sketch_id))
```
kiddinn, line 107: maybe add:

```
self.state.AddToCache('timesketch_sketch', sketch)
```

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -115,6 +133,10 @@ def Process(self):
     if not sketch:
```
kiddinn, line 133: maybe have this as

```
if not sketch and self.sketch_id:
```

since we kinda need that for this

kiddinn (commented):
some comments

kiddinn (approved):
LGTM

PR #343, opened 2020-10-15
Improvements to the SCP module

tomchop:
The SCP module can now SCP in both directions. This is a 2x increase in the directions SCP can copy files to.


dftimewolf/lib/exporters/scp_ex.py
```diff
@@ -48,25 +59,112 @@ def SetUp(self, # pylint: disable=arguments-differ
...
+    else:
+      # scp user@host:/path1 user@host:/path2 /destination
+      cmd.extend(self._PrefixRemotePaths(self._paths))
+      cmd.extend([self._destination])
+
+    self.logger.debug('Executing SCP command: {0:s}'.format(' '.join(cmd)))
+    ret = subprocess.call(cmd)
+    if ret != 0:
+      self.ModuleError(
+          'Failed copying {0!s}'.format(self._paths), critical=True)
```
kiddinn, line 103: why the `!s` why not `{0:s}...... format(', '.join(self._paths).....`

kiddinn (commented):
minor comments, basically just nits

dftimewolf/lib/exporters/scp_ex.py
tomchop: I'm really OK with the output being 

```
Failed copying ['path1', 'path2'] since it makes it clear what is part of the path and what isn't
```

I could do `{0:s}` and pass in `str(self._paths)` if you feel strongly about it.

kiddinn: do I feel strongly about it... no... not at all

kiddinn (commented):
massive nits... massive nits

kiddinn (approved):
LGTM

PR #345, opened 2020-10-19
Introduce runtime names for modules

tomchop:
Modules can now be identified by their `runtime_name` instead of their class names. This allows to disambiguate modules when there is more than one module of the same type in a recipe.

I didn't find a better variable name than `runtime_name`. WDYT?

### DIFF
--- a/dftimewolf/lib/module.py
+++ b/dftimewolf/lib/module.py
@@ -29,7 +29,6 @@ class BaseModule(object):
 
     Args:
       state (DFTimewolfState): recipe state.
-      name (Optional[str]): The module's runtime name.
       name (Optional[str]): A unique name for a specific instance of the module
           class. If not provided, will default to the module's class name.
       critical (Optional[bool]): True if the module is critical, which causes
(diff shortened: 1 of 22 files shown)


tomchop:
Moving to draft as I still need to add the kwarg to all existing modules. But feel free to take a look at the other changes.

dftimewolf/lib/module.py
```diff
@@ -19,25 +19,31 @@ class BaseModule(object):
...
   """
 
-  def __init__(self, state, critical=False):
+  def __init__(self, state, name=None, critical=False):
     """Initialize a module.
 
     Args:
       state (DFTimewolfState): recipe state.
+      name (Optional[str]): The module's runtime name.
+      name (Optional[str]): A unique name for a specific instance of the module
```
kiddinn, line 33: why is name twice here?

kiddinn (commented):
one tiny minor nit

dftimewolf/lib/module.py
tomchop: I was overeager to document this new feature it seems.

berggren (approved):
LGTM

PR #350, opened 2020-10-24
Turbinia artifacts

tomchop:


dftimewolf/lib/processors/turbinia_artifact.py
```diff
@@ -0,0 +1,97 @@
...
+          compressed_directory=fspath.path, local_path=fspath.path)
+      task_datas.append(self.TurbiniaProcess(evidence_))
+
+    self.logger.info('Files generated by Turbinia:')
+    for task_data in task_datas:
+      for task in task_data:
+        saved_paths = task.get('saved_paths') or []
+        for path in saved_paths:
+          # Ignore temporary files generated by turbinia
+          if path.startswith('/tmp'):
```
berggren, line 87: This is a bit fragile as this depends on how Turbinia is configured. If TMP_PATH in the turbinia conf i set to something else this will not match.

I'm not sure we can do much about that here, but worth knowing.

berggren: Oh, wait do you have access to the turbinia config here maybe? If that is the case you can get the config param from there

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -18,80 +18,41 @@
...
+  """Base class for processing with Turbinia.
 
   Attributes:
+    turbinia_config_file (str): Full path to the Turbinia config file to use.
     client (TurbiniaClient): Turbinia client.
-    disk_name (str): name of the disk to process.
     instance (str): name of the Turbinia instance
-    project (str): name of the GPC project containing the disk to process.
+    project (str): name of the GCP project containing the disk to process.
+    run_all_jobs (bool): Whether to run all jobs instead of a faster subset.
```
berggren, line 30: What jobs does this include? It would be nice to be able to pass through a job whitelist to Turbinia so one can choose (as an advanced feature). 

In any case, let's be clear what jobs are included/excluded here.

berggren (changes requested):
Looks good, just a few comments

dftimewolf/lib/processors/turbinia_gcp.py
tomchop: We only exclude slow-running jobs here ('StringsJob', 'BinaryExtractorJob', 'BulkExtractorJob', 'PhotorecJob'), added to the docstring. I guess I'll never have the full list of jobs that are actually run because this is dependent on the Turbinia deployment.

I think it would be better to rename this parameter into `skip_slow_jobs`. I opened https://github.com/log2timeline/dftimewolf/issues/351 to track this.

dftimewolf/lib/processors/turbinia_artifact.py
tomchop: Fixed.

dftimewolf/lib/exporters/local_filesystem.py
```diff
@@ -22,34 +23,57 @@ def __init__(self, state, name=None, critical=False):
...
+              file_container.path, self._target_directory)
+        except OSError as exception:
+          self.ModuleError(
+              'Could not copy files to {0:s}: {1!s}'.format(
+                  self._target_directory, exception),
+              critical=True)
+        for path_ in full_paths:
+          file_name = os.path.basename(path_)
+          self.state.StoreContainer(containers.File(name=file_name, path=path_))
+
```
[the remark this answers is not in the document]
tomchop: Fixed

dftimewolf/lib/processors/turbinia_gcp.py
berggren: 👍

berggren (approved):
LGTM

PR #358, opened 2020-11-09
Check target instance exists before launching analysis VM

ramo-j:
Pretty self explanatory I reckon. By checking that the target is valid before launching the VM, we don't launch the VM on a target error.

### DIFF
--- a/dftimewolf/lib/collectors/gcloud.py
+++ b/dftimewolf/lib/collectors/gcloud.py
@@ -3,6 +3,7 @@
 
 from google.auth.exceptions import DefaultCredentialsError, RefreshError
 from googleapiclient.errors import HttpError
+from libcloudforensics.errors import ResourceNotFoundError
 from libcloudforensics.providers.gcp.internal import project as gcp_project
 from libcloudforensics.providers.gcp import forensics as gcp_forensics
 
@@ -159,6 +160,16 @@ class GoogleCloudCollector(module.BaseModule):
             name=self._ANALYSIS_VM_CONTAINER_ATTRIBUTE_NAME,
             type_=self._ANALYSIS_VM_CONTAINER_ATTRIBUTE_TYPE,
             value=analysis_vm_name))
+
+    try:
+      if self.remote_instance_name:
+        self.remote_project.compute.GetInstance(self.remote_instance_name)
+    except ResourceNotFoundError as exception:
+      self.ModuleError(
+        message='Instance "{0:s}" not found or insufficient permissions'.format(
+          self.remote_instance_name),
+        critical=True)
+      return
 
     try:
       # TODO: Make creating an analysis VM optional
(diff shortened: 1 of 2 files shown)


dftimewolf/lib/collectors/gcloud.py
```diff
@@ -161,6 +162,9 @@ def SetUp(self,
             value=analysis_vm_name))
 
     try:
+      if self.remote_instance_name:
+        self.remote_project.compute.GetInstance(self.remote_instance_name)
+
```
tomchop, line 167: it's always better to make the try / except block as small as possible around the code that can actually raise the exception. in this case, binging up the block in L183 to L167:

```python
if self.remote_instance_name:
  try:
    self.remote_project.compute.GetInstance()
  except ResourceNotFoundError as exception:
    #...
```

Also, I would move this whole block to be one of the first functions to be executed in SetUp (so we don't create the analysis VM for nothing if we don't have access to the target instance or if it doesn't exist).

Finally, I'd extract this to its own function within the module (and to be frank this whole SetUp function could use some refactoring - for a future PR maybe)

tomchop (changes requested):

tests/lib/collectors/gcloud.py
```diff
@@ -51,16 +51,19 @@ def testInitialization(self):
...
+  @mock.patch('libcloudforensics.providers.gcp.internal.compute.GoogleCloudCompute.GetInstance')
   @mock.patch('libcloudforensics.providers.gcp.internal.compute_base_resource.GoogleComputeBaseResource.AddLabels')
   @mock.patch('libcloudforensics.providers.gcp.internal.compute_base_resource.GoogleComputeBaseResource')
   @mock.patch('libcloudforensics.providers.gcp.forensics.StartAnalysisVm')
   def testSetUp(self,
                 mock_StartAnalysisVm,
                 mock_GoogleComputeBaseResource,
-                mock_AddLabels):
+                mock_AddLabels,
+                mock_TargetInstance):
```
tomchop, line 62: nit: to be consistent with other mocks, can we name the mock variables after the function they're mocking? In this case

```suggestion
                mock_GetInstance):
```

Same for the variable in Line 117

tests/lib/collectors/gcloud.py
```diff
@@ -109,11 +113,13 @@ def testProcess(self,
                   mock_CreateDiskCopy,
                   mock_StartAnalysisVm,
                   mock_AddLabels,
-                  mock_GetBootDisk):
+                  mock_GetBootDisk,
+                  mock_TargetInstance):
```
tomchop, line 117:
```suggestion
                  mock_GetInstance):
```

tomchop (changes requested):
A nit, but otherwise LGTM!

PR #363, opened 2020-11-17
Created a SanityChecks preflight module

ramo-j:
Creation of a preflight module designed to fail fast for improper parameters.

Inital tests are that a start date is before an end date, and that the dates match an expected format.

Usage involves adding the module to a recipe and passing in the parameters you want tested, for example:

```
  "preflights": [
    ....,
    {
    "wants":[],
    "name": "SanityChecks",
    "args": {
      "startdate": "@start_date",
      "enddate": "@end_date",
      "dateformat": "%Y-%m-%d"
      }
    }
  ],
```


tests/lib/preflights/sanity_checks.py
```diff
@@ -0,0 +1,63 @@
...
+    """Test that dates in the incorrect order produces an error."""
+    test_state = state.DFTimewolfState(config.Config)
+    checker = sanity_checks.SanityChecks(test_state)
+
+    checker.SetUp(
+        startdate='2020-10-31', enddate='2020-10-01', dateformat='%Y-%m-%d')
+    checker.Process()
+
+    # The incorrect date order should mean an error was encountered
+    self.assertEqual(
```
tomchop, line 31: For each of these tests, can we also check that the error message is the one we expect?

dftimewolf/lib/preflights/sanity_checks.py
```diff
@@ -0,0 +1,62 @@
...
+      end = datetime.datetime.strptime(self.enddate, self.dateformat)
+      if start > end:
+        self.ModuleError(
+            # pylint: disable=line-too-long
+            'Hold fast time traveller, your start date {0:s} is after your end date {1:s}'
+            .format(self.startdate, self.enddate),
+            critical=True)
+    except (ValueError) as e:  # Date parsing failure
+      self.ModuleError(str(e), critical=True)
+
```
tomchop, line 57: According to the [l2t style guide](https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#exceptions):

```suggestion
    except (ValueError) as exception:  # Date parsing failure
      self.ModuleError(str(exception), critical=True)
```

(this probably needs to be changed in other parts of the code as well 😅)

dftimewolf/lib/preflights/sanity_checks.py
```diff
@@ -0,0 +1,62 @@
...
+    self.dateformat = dateformat
+
+  def Process(self):
+    """Test whether the values we've received are sane."""
+
+    try:
+      if (self.startdate and self.enddate and self.dateformat):
+        self._AreDatesValid()
+    except DFTimewolfError:  # We don't need the extra stacktrace here
+      return
```
tomchop, line 41: If you catch the `DFTimewolfError` here, as far as the scheduler is concerned the module will be marked as "successfully completed". We want the exception to propagate up to the state where it will [be caught](https://github.com/log2timeline/dftimewolf/blob/master/dftimewolf/lib/state.py#L175) and the recipe flow will be interrupted. So no need to catch it here.

tomchop (changes requested):

dftimewolf/lib/preflights/sanity_checks.py
ramo-j: Yeah, duly noted, and updated. It wold be nice if there was a way to catch the logic error, but not print the stacktrace in cases this module correctly handles.

dftimewolf/lib/preflights/sanity_checks.py
ramo-j: Done.

tests/lib/preflights/sanity_checks.py
ramo-j: Done.

dftimewolf/lib/preflights/sanity_checks.py
tomchop: Well you can raise any other error (`RuntimeError` being quite generic but easy enough to add) in your function and catch that one.

tests/lib/preflights/sanity_checks.py
```diff
@@ -28,37 +29,37 @@ def testInvalidDateOrder(self):
...
-        'Date order validation succeeded when it should have failed')
-    self.assertEqual(
-        checker.state.errors[0].message,
-        expected_error,
-        'Error message differs from expected')
+    try:
+      checker.SetUp(
+          startdate=start_date, enddate=end_date, dateformat='%Y-%m-%d')
+      checker.Process()
+    except (DFTimewolfError) as exception:
```
tomchop, line 36: Please use `assertRaises` here: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/lib/preflights/sanity_checks.py
```diff
@@ -28,37 +29,37 @@ def testInvalidDateOrder(self):
...
     checker = sanity_checks.SanityChecks(test_state)
 
-    checker.SetUp(
-        startdate='20-10-01', enddate='20-10-31', dateformat='%Y-%m-%d')
-    checker.Process()
+    try:
+      checker.SetUp(
+          startdate='20-10-01', enddate='20-10-31', dateformat='%Y-%m-%d')
+      checker.Process()
+    except (DFTimewolfError) as exception:
```
tomchop, line 49: Please use `assertRaises` here: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/lib/preflights/sanity_checks.py
```diff
@@ -0,0 +1,69 @@
...
+
+  def testInvalidDateOrder(self):
+    """Test that dates in the incorrect order produces an error."""
+    test_state = state.DFTimewolfState(config.Config)
+    checker = sanity_checks.SanityChecks(test_state)
+
+    start_date = '2020-10-31'
+    end_date = '2020-10-01'
+    expected_error = sanity_checks.DATE_ORDER_ERROR_STRING.format(
+        start_date, end_date)
```
tomchop, line 30: During tests, it's always better to use as literal values as possible, to limit the amount of logic in the test (even if it's simple) but also to make the test very clear to whoever's reading it. (just like you did in `testInvalidDateFormat`)

I would just call `SetUp` with hardcoded params: `SetUp(startdate='2020-10-31', end_date='2020-10-01', dateformat=...)` and compare the resulting exception with the appropriate string literal.

tomchop (changes requested):

tests/lib/preflights/sanity_checks.py
ramo-j: Done.

tests/lib/preflights/sanity_checks.py
ramo-j: Done.

tests/lib/preflights/sanity_checks.py
ramo-j: Done.

tomchop (approved):
LGTM, thanks!

PR #366, opened 2020-11-20
Make search by FQDN fail if multiple recent results return

sa3eed3ed:
Make search by FQDN fail if multiple results with clients seen in the… last month return or if multiple results, where non of them is seen in the last month return.
This will help avoid picking up on the wrong client.


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -103,6 +103,39 @@ def _GetClientBySelector(self, selector):
...
+        boolean: True if the timestamp is in last month from now.
+      """
+      last_seen_datetime = datetime.datetime.utcfromtimestamp(
+          timestamp / 1000000)
+      diff_year_in_month = (datetime.datetime.utcnow().year
+          - last_seen_datetime.year) * 12
+      diff_month_in_month = (datetime.datetime.utcnow().month
+          - last_seen_datetime.month)
+      diff_month = diff_year_in_month + diff_month_in_month
+      return bool(diff_month < 1)
```
tomchop, line 122: A simpler way to do this is to do a check like:

```
month_ago = datetime.datetime.utcnow() - datetime.timedelta(30)  # 30 days before now()
return  last_seen_datetime > month_ago
```

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -103,6 +103,39 @@ def _GetClientBySelector(self, selector):
...
+
+    last_month_list = list(filter(lambda x: _SeenLastMonth(x[0]), result))
```
tomchop, line 124: I would rather move this logic to a separate function that takes a list of clients (`result`) and returns only the ones that were active in the past month (along with the logic in `_SeenLastMonth`)

If the list is empty, we should output an error message saying that no matching clients have been online for the past month. If the list is >1, then output the error message as below (also maybe stating how many clients exactly were returned by the query?)

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -103,6 +103,39 @@ def _GetClientBySelector(self, selector):
...
+            'Please use client ID instead of the hostname.'.format(
+        selector), critical=True)
+    if not last_month_list and len(result) > 1:
+      self.ModuleError(
+            'Multiple hosts with the same FQDN: "{0:s}" non '
+            'of them have been active in the last month.\n'
+            'Please use client ID instead of the hostname.'.format(
+        selector), critical=True)
+
     last_seen, client = sorted(result, key=lambda x: x[0], reverse=True)[0]
```
tomchop, line 139: This line would go into the function that filters clients to only the relevant ones, so we can remove it from here.

tomchop (changes requested):

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: Done

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: Done

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: Done

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -69,6 +69,33 @@ def SetUp(
...
+        timestamp / 1000000)
+    # 30 days before now()
+    month_ago = datetime.datetime.utcnow() - datetime.timedelta(30)
+    return  last_seen_datetime > month_ago
+
+  def _ActiveLastMonth(self, result):
+    """Take a list of clients and return clients active last month.
+
+    Args:
+      result list[str]: A list of clients.
```
tomchop, line 91: `result` here is not a `list[str]`, but a list of tuples, `[0]` being the last seen timestamp and `[1]` being the client object. (have a look at L127)

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -69,6 +69,33 @@ def SetUp(
...
+    Returns:
+      boolean: True if the timestamp is in last month from now.
+    """
+    last_seen_datetime = datetime.datetime.utcfromtimestamp(
+        timestamp / 1000000)
+    # 30 days before now()
+    month_ago = datetime.datetime.utcnow() - datetime.timedelta(30)
+    return  last_seen_datetime > month_ago
+
+  def _ActiveLastMonth(self, result):
```
tomchop, line 87: I'd rename this to `_FilterActiveClients`

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -103,7 +130,21 @@ def _GetClientBySelector(self, selector):
       self.ModuleError('Could not get client for {0:s}'.format(
           selector), critical=True)
 
-    last_seen, client = sorted(result, key=lambda x: x[0], reverse=True)[0]
+    last_month_list = self._ActiveLastMonth(result)
```
tomchop, line 133:
```suggestion
    active_clients = self._ActiveLastMonth(result)
```

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -103,7 +130,21 @@ def _GetClientBySelector(self, selector):
...
+            'Multiple hosts ({0:d}) with the same '
+            'FQDN: "{1:s}" have been active in the last month.\n'
+            'Please use client ID instead of the hostname.'.format(
+                len(last_month_list), selector), critical=True)
+    if not last_month_list:
+      self.ModuleError(
+            'No host with FQDN: "{0:s}" '
+            'has been active in the last month.\n'
+            'Please use client ID instead of the hostname.'.format(
+                selector), critical=True)
```
tomchop, line 145: This error message will be displayed to the user if no active clients can be found, regardless of the selector (FQDN or client_id), which will be a bit confusing if the user is already using a Client ID. A better message would be `no 30-day active client found for selector {selector}`

Additionally, it would probably be interesting to specify how many old / inactive clients were found, so that the user can instantly get an idea of whether they typed the FQDN wrong or if Clients were found but not active.

tomchop (changes requested):

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: done

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: done

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: done

dftimewolf/lib/collectors/grr_hosts.py
sa3eed3ed: done

dftimewolf/lib/collectors/grr_hosts.py
tomchop: You forgot to change the docstring in the `Args:` section

tomchop (dismissed):
Small nit about the docstring, otherwise LGTM

tomchop (approved):

PR #371, opened 2021-02-09
Fix for audit log collecting

ramo-j:
Some audit logs currently fail to successfully be retrieved and processed in the `GCPLogsCollector` module. This PR updates:

* the auto generated protobuf for audit logs
* the requirements.txt for the same, and
* the `_CustomToAPIRepr` monkey patch

to fix this issue.


tomchop:
https://github.com/log2timeline/dftimewolf/pull/376 should get rid of the issue that this fixes, since we're not using GRPC anymore but just JSON. Can you confirm that you can't reproduce this anymore?

ramo-j:
Yep, that PR looks to fix the issue. Closing this one out.

PR #389, opened 2021-05-05
Hosts in GRR modules are attribute containers instead of strings

tomchop:
This PR changes the way GRR modules deal with hosts. Instead of strings, they now manipulate `containers.Host` attribute containers. This makes it possible for other modules (e.g. "user2host") to generate containers for the GRR modules to work on.

### DIFF
--- a/dftimewolf/lib/collectors/grr_hosts.py
+++ b/dftimewolf/lib/collectors/grr_hosts.py
@@ -330,7 +330,7 @@ class GRRArtifactCollector(GRRFlow):
   Attributes:
     artifacts (list[str]): artifact definition names.
     extra_artifacts (list[str]): extra artifact definition names.
-    hostnames (list[containers.Host]): Hosts to collect artifacts from.
+    hosts (list[containers.Host]): Hosts to collect artifacts from.
     use_tsk (bool): True if GRR should use Sleuthkit (TSK) to collect file
         system artifacts.
   """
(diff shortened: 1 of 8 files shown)


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -404,7 +404,12 @@ def SetUp(self,
       self.extra_artifacts = [item.strip() for item
                               in extra_artifacts.strip().split(',')]
 
-    self.hostnames = [item.strip() for item in hosts.strip().split(',')]
+    for item in hosts.strip().split(','):
```
Onager, line 407: How about renaming "hosts" here to "hostnames"? This is a list of hostnames, not host containers

dftimewolf/lib/containers/containers.py
```diff
@@ -223,3 +223,19 @@ def __init__(self, data_frame, description, name):
...
+  """Attribute container definition for a host.
+
+  Attributes:
+    hostname (str): The host's hostname.
+    platform (str): The host's platform. One of {win, linux, macos, unknown}.
+  """
+
+  CONTAINER_TYPE = 'host'
+
+  def __init__(self, hostname, platform='unkown'):
```
Onager, line 238: "unknown"

Onager (changes requested):

Onager (approved):

PR #390, opened 2021-05-05
Better module import

tomchop:
This change imports modules dynamically - only the modules declared in the recipe will be imported at runtime.


giovannt0 (approved):
LGTM

Side nit, is there a preferred string formatting method in use in this repo? I see usage of both f'xxx {yyy} zzz' and 'xxx {0:s} zzz'.format(yyy)

tomchop:
That's a good point. I opened https://github.com/log2timeline/dftimewolf/issues/391 to move the discussion there.

PR #404, opened 2021-06-15
Fix misc issues with GCP logging collector and processor

Onager:
Better handling for quota limits and better preservation of some log information in retrieved logs.


tomchop (approved):
Thanks! I wonder what the overlap with cloud-foreniscs-utils lib is here. I'll open an issue to see if leveraging it here is a viable solution.

PR #413, opened 2021-08-04
Add test to check that modules are invoked with the expected parameters.

tomchop:


data/recipes/upload_web_ts.json
```diff
@@ -15,19 +15,20 @@
...
-	      "analyzers": "browser_search,browser_timeframe,account_finder,phishy_domains,evtx_gap,login,win_crash,safebrowsing,chain,"
+            "wait_for_timelines": "@wait_for_timelines",
+	        "analyzers": "browser_search,browser_timeframe,account_finder,phishy_domains,evtx_gap,login,win_crash,safebrowsing,chain,"
         }
     }, {
         "wants": ["TimesketchExporter"],
         "name": "TimesketchEnhancer",
         "args": {
-	    "wait_for_analyzers": "@wait_for_analyzers",
+	        "wait_for_analyzers": "@wait_for_analyzers",
```
giovannt0, line 25: Formatting looks odd here?

dftimewolf/lib/collectors/azure.py
```diff
@@ -40,7 +40,10 @@ class AzureCollector(module.BaseModule):
   _ANALYSIS_VM_CONTAINER_ATTRIBUTE_NAME = 'Analysis VM'
   _ANALYSIS_VM_CONTAINER_ATTRIBUTE_TYPE = 'text'
 
-  def __init__(self, state: DFTimewolfState, critical: bool=False) -> None:
+  def __init__(self,
+               state: DFTimewolfState,
+               name: Optional[str]=None,
```
giovannt0, line 45: `name` is missing from the docstring.

tests/cli/main_tool.py
```diff
@@ -36,3 +38,28 @@ def testToolWithArbitraryRecipe(self):
...
+    for recipe in tool._recipes_manager.GetRecipes():
+      tool._state.LoadRecipe(recipe.contents, dftimewolf_recipes.MODULES)
+      for module in recipe.contents['modules']:
+        runtime_name = module.get('runtime_name', module['name'])
+        setup_func = tool.state._module_pool[runtime_name].SetUp
+        expected_args = set(inspect.getfullargspec(setup_func).args)
+        expected_args.remove('self')
+        provided_args = set(module['args'])
+
+        self.assertEqual(
```
giovannt0, line 62: Do we need to be concerned by argument order?
Say, I have a recipe that takes `--arg1` and `--arg2`. Due to `--`, both are optional. If I pass `dftimewolf myrecipe --arg1=foo --arg2=bar` or `dftimewolf myrecipe --arg2=bar --arg1=foo` will it work?

giovannt0 (approved):
Couple nits, otherwise LGTM

tests/cli/main_tool.py
tomchop: They all end up being passed as kwargs, so the order doesn't really matter so long as the names match.

Now that all recipes are passing all parameters (optional or not), a good question to ask is whether we still want positional vs. keyword parameters in the module's SetUp() functions, or if we should just consistently have them be positional, and specify the default value in the recipe. (I would favor the latter)

giovannt0: The latter SGTM

PR #425, opened 2021-08-18
Add custom logging class to be able to log "success" messages

tomchop:
Logging "success" messages allows for module writers to highlight important messages in their modules such as successful download destinations, new timesketch sketch URLs, etc.


giovannt0 (dismissed):
VERY NICE

PR #432, opened 2021-09-07
Threaded Modules for TurbiniaGCPProcessor and TimesketchExporter

ramo-j:
Created two new modules. Both are copies of existing modules, `TurbiniaGCPProcessor` and `TimesketchExporter`, but they use the ThreadedAwareModule for the base class. Some adjustments needed to be made to allow this to operate properly: For example, `TurbiniaGCPProcessor` sourced a disk from previous modules using the `containers.ForensicsVM` container. Created a new container specifically for GCE Disks, and used that instead.

Also, made some adjustments to the threaded module process - namely, made most of the static methods non-static, as this makes more sense given the eventual direction of threaded module handling, that only attempting an implementation revealed.


itsmvd (changes requested):

tests/test_modules/thread_aware_modules.py
```diff
@@ -94,22 +94,17 @@ def Process(self, container) -> None:
   def GetThreadOnContainerType():
     return TestContainer
 
-  @staticmethod
-  def GetThreadPoolSize():
+  def GetThreadPoolSize(self):
     return 2
```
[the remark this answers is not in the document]
ramo-j: The test runs three threads, I thought it made sense to have the simultaneous limit be less than the total.

dftimewolf/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,248 @@
+# -*- coding: utf-8 -*-
+"""Export processing results to Timesketch."""
```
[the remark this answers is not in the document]
ramo-j: Done.

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,358 @@
+# -*- coding: utf-8 -*-
+"""Processes GCP cloud disks using Turbinia."""
```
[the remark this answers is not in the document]
ramo-j: Done.

data/recipes/gcp_turbinia_ts_threaded.json
```diff
@@ -0,0 +1,44 @@
+{
+    "name": "gcp_turbinia_ts_threaded",
+    "description": "Process GCP persistent disks with Turbinia and send output to Timesketch.\n\nThis processes disks that is already in the project where Turbinia exists. If you want to copy the disk from another project, use the gcp_turbinia_disk_copy_ts recipe.",
```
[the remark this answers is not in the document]
ramo-j: Done.

ramo-j:
As per OOB discussion, tagging @Onager to have a look at this as well.

dftimewolf/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,250 @@
...
+            'No write access to sketch ID {0:d}, aborting'.format(
+                self.sketch_id),
+            critical=True)
+      self.state.AddToCache('timesketch_sketch', self.sketch)
+      self.sketch_id = self.sketch.id
+
+    if analyzers and isinstance(analyzers, (tuple, list)):
+      self._analyzers = analyzers
+
+  def _CreateSketch(self, incident_id: None=None) -> ts_sketch.Sketch:
```
giovannt0, line 108:
```suggestion
  def _CreateSketch(self, incident_id: Optional[str] = None) -> ts_sketch.Sketch:
```

dftimewolf/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,250 @@
...
+    for attribute in attributes:
+      if attribute.name == self._SKETCH_ATTRIBUTE_NAME:
+        sketch_match = re.search(r'sketch/(\d+)/', attribute.value)
+        if sketch_match:
+          sketch_id = int(sketch_match.group(1), 10)
+          return sketch_id
+    return 0
+
+  def Process(self, container: containers.File) -> None:
+    """Executes a Timesketch export."""
```
giovannt0, line 160: Docstring missing for arguments

dftimewolf/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,250 @@
...
+
+    recipe_name = self.state.recipe.get('name', 'no_recipe')
+
+    description = container.name
```
giovannt0, line 164: Looks like you're using both description and container.name below, maybe use just one?

dftimewolf/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,250 @@
...
+        continue
+      self.logger.info('Analyzer: {0:s} is running, session ID: {1:d}'.format(
+          analyzer, session_id))
+      self.logger.info(results.status_string)
+
+  @staticmethod
+  def GetThreadOnContainerType() -> Type[interface.AttributeContainer]:
+    return containers.File
+
+  def GetThreadPoolSize(self) -> int:
```
giovannt0, line 218: Is this really worth to have as a method? Since the returned value is always the same, might as well have a constant? Or is this an override of the parent class? In which case just ignore me

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+
+    Returns:
+      tuple[list, list]: A tuple of two lists. The first element contains the
+          local paths, the second element contains the remote (GS) paths.
+    """
+    local_paths = []
+    gs_paths = []
+    for task in task_data:
+      # saved_paths may be set to None
+      saved_paths = task.get('saved_paths') or []
```
giovannt0, line 86:
```suggestion
      saved_paths = task.get('saved_paths', [])
```

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+                            gs_paths: List[str]) -> List[Tuple[str, str]]:
+    """Downloads files stored in Google Cloud Storage to the local filesystem.
+
+    Args:
+      timeline_label (str): Label to use to construct the path list.
+      gs_paths (List[str]):  gs:// URI to files that need to be downloaded
+          from GS.
+
+    Returns:
+      list:
```
giovannt0, line 111: This may trigger some weird things on auto generated doc. Perhaps `list[tuple[str, str]]`: Tuple containing....

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+    self.sketch_id = sketch_id
+    self.run_all_jobs = run_all_jobs
+
+    turbinia_config.LoadConfig(config_file=self.turbinia_config_file)
+    if not self.project:
+      self.project = turbinia_config.TURBINIA_PROJECT
+    if not self.turbinia_zone:
+      self.turbinia_zone = turbinia_config.TURBINIA_ZONE
+
+    if self.project is None or self.turbinia_zone is None:
```
giovannt0, line 160: If I'm reading this correctly, the empty string will skip that safeguard. However seeing the error that's thrown below it sounds like an empty string would also break things. Is that correct? If so, consider using `if not self.project or not self.turbinia_zone` instead.

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+      return
+    self._output_path = tempfile.mkdtemp()
+    self.client = turbinia_client.get_turbinia_client(run_local=False)
+
+  def TurbiniaProcess(
+      self, evidence_: evidence.Evidence) -> List[Dict[str, Any]]:
+    """Creates and sends a Turbinia processing request.
+
+    Args:
+      evidence_(turbinia.evidence.Evidence): The evience to process
```
giovannt0, line 183:
```suggestion
      evidence_(turbinia.evidence.Evidence): The evidence to process
```

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+    task_data = []  # type: List[Dict[str, str]]
+    try:
+      self.logger.success(
+          'Creating Turbinia request {0:s} with Evidence {1!s}'.format(
+              request.request_id, evidence_.name))
+      self.client.send_request(request)
+      self.logger.info('Waiting for Turbinia request {0:s} to complete'.format(
+          request.request_id))
+      self.client.wait_for_request(**request_dict)
+      task_data = self.client.get_task_data(**request_dict)
```
giovannt0, line 225: Can we move the logger statements outside the try block?

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+
+    try:
+      self.TurbiniaSetUp(project, turbinia_zone, sketch_id, run_all_jobs)
+    except TurbiniaException as exception:
+      self.ModuleError(str(exception), critical=True)
+      return
+
+  # pylint: disable=arguments-renamed
+  def Process(self, disk_container: containers.GCEDisk) -> None:
+    """Process a disk with Turbinia."""
```
giovannt0, line 295: Docstring is missing argument

dftimewolf/lib/processors/turbinia_gcp_tam.py
```diff
@@ -0,0 +1,361 @@
...
+      if path.endswith('.plaso'):
+        self.logger.success('Found plaso result: {0:s}'.format(path))
+        container = containers.File(name=description, path=path)
+      self.state.StoreContainer(container)
+
+  @staticmethod
+  def GetThreadOnContainerType() -> Type[interface.AttributeContainer]:
+    return containers.GCEDisk
+
+  def GetThreadPoolSize(self) -> int:
```
giovannt0, line 345: Well, seeing it here again I'm guessing this is an override of the parent class. I'm wondering if it would be a good idea to move that `5` value to a constant defined somewhere (perhaps a shared config file if that already exists) so that we can easily adapt this in the future.

tests/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,144 @@
...
+    timesketch_exporter = timesketch.TimesketchExporterThreaded(test_state)
+    self.assertIsNotNone(timesketch_exporter)
+
+  # pylint: disable=invalid-name
+  @mock.patch('dftimewolf.lib.timesketch_utils.GetApiClient')
+  def testSetupWithSketchId(self, mock_GetApiClient):
+    """Tests the SetUp function."""
+    mock_sketch = mock.Mock()
+    mock_sketch.id = 1234
+    mock_sketch.my_acl = ['write']
```
giovannt0, line 31: Isn't `mock.Mock()` taking arbitrary arguments such that you can inline mocks? E.g. `mock.Mock(id=1234, my_acl=['write'])`. Note that I remember seeing some flaky behaviour when experimenting with those, but perhaps in your case it'd work and reduce a bit the code.

tests/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,144 @@
...
+    mock_api_client.get_sketch.return_value = mock_sketch
+    mock_GetApiClient.return_value = mock_api_client
+    test_state = state.DFTimewolfState(config.Config)
+    timesketch_exporter = timesketch.TimesketchExporterThreaded(test_state)
+    timesketch_exporter.SetUp(
+        incident_id=None,
+        sketch_id=1234,
+        analyzers=None
+    )
+    self.assertEqual(timesketch_exporter.sketch_id, 1234)
```
giovannt0, line 42: Mega-nit: self.assertEqual's first argument is the expected value AFAIK, so this should be inverted to `self.assertEqual(1234, ...)`

tests/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,144 @@
...
+    test_state = state.DFTimewolfState(config.Config)
+    timesketch_exporter = timesketch.TimesketchExporterThreaded(test_state)
+    with self.assertRaises(errors.DFTimewolfError) as error:
+      timesketch_exporter.SetUp(
+          incident_id=None,
+          sketch_id=1234,
+          analyzers=None
+      )
+
+    self.assertEqual(
```
giovannt0, line 64: Same as above, and same for below assertions

giovannt0 (changes requested):

dftimewolf/lib/exporters/timesketch_tam.py
```diff
@@ -0,0 +1,250 @@
...
+      analyzers (Optional[List[str]): If provided a list of analyzer names
+          to run on the sketch after they've been imported to Timesketch.
+      token_password (str): optional password used to decrypt the
+          Timesketch credential storage. Defaults to an empty string since
+          the upstream library expects a string value. An empty string means
+          a password will be generated by the upstream library.
+      wait_for_timelines (bool): Whether to wait until timelines are processed
+          in the Timesketch server or not.
+    """
+    self.wait_for_timelines = bool(wait_for_timelines)
```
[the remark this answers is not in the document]
ramo-j: My excuse for a lot of these is going to be "It was copy pasted from the source module :)"

dftimewolf/lib/exporters/timesketch_tam.py
ramo-j: Yep this is a method to override from the parent class. The value return needs to be determined by the implementation - Though to be fair, I put no effort into working out what the "correct" value here should be.

dftimewolf/lib/processors/turbinia_gcp_tam.py
ramo-j: As mentioned for the other module - The value here is dependant on what the threaded module is actually doing, and is up to the implementation to decide what size thread pool it should support.

dftimewolf/lib/processors/turbinia_gcp_tam.py
ramo-j: We've got a logging statement either side of `send_request` so would require 2 try/except blocks, so I'd prefer not to fo it looks a bit cleaner.

tests/lib/exporters/timesketch_tam.py
ramo-j: Hmm the unittest docs [1] only list the args as `first` and `second` as opposed to expected and actual.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

giovannt0 (approved):
LGTM

itsmvd (approved):

PR #447, opened 2021-10-20
Bigquery collector and related recipes

anttitikkanen:


tests/lib/collectors/bigquery.py
```diff
@@ -0,0 +1,33 @@
...
+
+  @mock.patch('google.cloud.bigquery.Client')
+  def testQuery(self, mock_bq):
+    """Tests that the collector calls the BQ client."""
+    mock_bq().query().to_dataframe().to_json.return_value = "{'foo':1}"
+    test_state = state.DFTimewolfState(config.Config)
+    bq_collector = bigquery.BigQueryCollector(test_state)
+    bq_collector.SetUp('test_project', 'test_query', 'test_description')
+    bq_collector.Process()
+    mock_bq().query().to_dataframe().to_json.assert_called_once()
```
tomchop, line 30: Could we also make sure that `quer()` is being called with the expected parameters (in this case `'test_query'`)?

PR #458, opened 2021-11-02
Refactored GRR modules

ramo-j:
Refactoring of grr modules to take advantage of parallel processing. Additionally, added support for multiple flows to `GRRFlowCollector` which previously supported only one flow input.


dftimewolf/lib/containers/containers.py
```diff
@@ -353,6 +353,22 @@ def __init__(self, hostname: str, platform: str = 'unknown') -> None:
     self.platform = platform
 
 
+class GrrHostFlowPair(interface.AttributeContainer):
```
tomchop, line 356: Given that Flows are always associated with a Host (as they are unique *per host*), I think we can just rename this to `GrrHostFlow` or even `GrrFlow`.

dftimewolf/lib/collectors/grr_base.py
```diff
@@ -107,20 +104,20 @@ def _WrapGRRRequestWithApproval(
...
+          logger.success(approval_url) # type: ignore
           time.sleep(self._CHECK_APPROVAL_INTERVAL_SEC)
           continue
 
         # If no approvers were specified, abort.
         if not self.approvers:
           message = ('GRR needs approval but no approvers specified '
                      '(hint: use --approvers)')
-          self.ModuleError(message, critical=True)
+          raise GrrError(message, critical=True) from exception
```
tomchop, line 120: Why not simply raise a `dfTimewolfError` here? You're not catching any `GrrError`s anyways.

tomchop (changes requested):
Some minor changes, otherwise this looks good to me!

dftimewolf/lib/collectors/grr_base.py
ramo-j: `GrrError` is a subclass of `DFTimewolfError`. Added in `Raises` notations.

tomchop: I meant, why introduce a new `GrrError` class, when dfTimewolf error seems to be enough for this purpose? I don't see anywhere in the code where you're doing `except GrrError as error` (i.e. catching that specific error type)

ramo-j: Good point. Removed `GrrError` in favour of `DFTimewolfError`.

PR #470, opened 2021-11-25
Type fix

ramo-j:
Small correction to a type annotation.


itsmvd (approved):

ramo-j:
@tomchop looks like we can't merge without you these days with the two build tests that don't finish.

tomchop:
The two build things never finishing is super annoying, I'll look into it. Thanks for the fix!

PR #472, opened 2021-11-29
Add option to stop target instance

dfjxs:
Added an option to the GCP collection recipes to also stop the target instance after collecting disks


dftimewolf/lib/collectors/gcloud.py
```diff
@@ -89,6 +92,24 @@ def Process(self) -> None:
...
+              self.remote_instance_name)
+          # TODO(dfjxs): Account for GKE Nodes
+          remote_instance.Stop()
+        except RuntimeError as exception:
+          self.ModuleError(str(exception), critical=False)
+        self.logger.success(
+            'Stopped instance {0:s}'.format(self.remote_instance_name))
+      else:
+        self.logger.warning(
+            'Could not stop instance, instance name not provided')
```
tomchop, line 108: Should we add this warning to the SetUp phase? Maybe even bail if the user provides that flag and no instance name?

dftimewolf/lib/collectors/gcloud.py
```diff
@@ -89,6 +92,24 @@ def Process(self) -> None:
...
       self.state.StoreContainer(container)
 
+    if self.stop_instance and not warned:
+      if self.remote_instance_name:
+        try:
+          remote_instance = self.remote_project.compute.GetInstance(
+              self.remote_instance_name)
+          # TODO(dfjxs): Account for GKE Nodes
+          remote_instance.Stop()
+        except RuntimeError as exception:
```
tomchop, line 102: Can we catch a narrower exception here?

tomchop (changes requested):

dftimewolf/lib/collectors/gcloud.py
dfjxs: Ah, good point. Will do.

dftimewolf/lib/collectors/gcloud.py
dfjxs: Updated.

tomchop (approved):

PR #483, opened 2021-12-07
Automatic documentation generation

### LINKED ISSUE #476: Auto-generate recipe markdown documentation
Parse recipes, auto-generate markdown documentation from recipe medatada.
- [ ] Markdown docs are generated for each recipe in a `recipes/recipe_name.md` file
- [ ] Generate a module chaining graph for each recipe
- [ ] This generation is run on push to master, along with the doc build

tomchop:
This PR fixes #476 and adds automatic documentation generation to the dftimewolf dev workflow

* Generate markdown representations of recipes
* Generate a graphical representation of of module chaining
* Add github actions to run this at push time so documentaiton is up-to-date.

Sample of generated `recipe-list.md`:

image

To generate this file manually, run 

```
pipenv install --dev
python docs/generate_recipe_doc.py data/recipes
```


docs/generate_recipe_doc.py
```diff
@@ -0,0 +1,156 @@
...
+    m_cluster.attr(label='Modules', fontcolor='lightgrey')
+    m_cluster.attr(fontname='Sans')
+    for module in recipe['modules']:
+      modname = module.get('runtime_name', module['name'])
+      m_cluster.node(modname, shape='box', fontname='Sans')
+      for wants in module['wants']:
+        m_cluster.edge(wants, modname)
+
+  if recipe.get('preflights'):
+    first_module = recipe['preflights'][0]
```
ramo-j, line 121: How will this behave if someone includes an empty `preflight` section in a recipe?

```
{
  "name": "name",
  "preflights": [],
  "modules": [.....
```

Same question for the `modules` section on line 124, but chances are far less that section will be empty.

docs/generate_recipe_doc.py
```diff
@@ -0,0 +1,156 @@
...
+
+```
+pipenv install --dev
+python docs/generate_recipe_doc.py data/recipes
+```
+
+---
+"""
+
+# This is the template for representing a recipe's documentation
```
ramo-j, line 26: Comments left in for a reason?

tomchop: I guess this was just to show how the template is intended to be used. I can remove it it you find it confusing.

docs/generate_recipe_doc.py
tomchop: If preflights are empty (or not specified), `recipe.get('preflights')` will evaluate to False, so this will be skipped, resulting in no "Preflight" box in the generated graph. 

There will always be a module in a recipe, which is why I just access `recipe['modules'][0]` in L124, so we should never encounter this case.

ramo-j (approved):

PR #487, opened 2021-12-09
Add tags to recipes

tomchop:


ramo-j:
I know this is a draft, but I saw this fly by and had a thought - What if modules had tags rather than recipes, and as such recipes gain tags from their constituent modules?

PR #492, opened 2022-01-06
Add GRR Hunt Client filters - by label and operating system

### LINKED ISSUE #493: Add Label/OS Client Filters to GRR Hunt collectors
It would be useful to apply client filters such as labels and OS type when running GRR hunts.

sydp:
Fixes #493


dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -54,8 +63,52 @@ def _CreateHunt(
...
+        label_rule.rule_type = label_rule.LABEL
+        label_rule.label.label_names.append(client_label)
+
+    if self.client_operating_systems:
+      client_operating_systems = set(
+          os for os in self.client_operating_systems.lower().split(',')
+          if os in ('win', 'osx', 'linux'))
+
+      if len(client_operating_systems) != len(self.client_operating_systems):
+        self.ModuleError('Invalid client OS values.', critical=True)
```
tomchop, line 90: Can we move this check to `SetUp()` ? This makes sure that a valid OS version is passed before anything else executes.

data/recipes/grr_hunt_file.json
```diff
@@ -26,6 +29,9 @@
         ["--verify", "Whether to verify the GRR TLS certificate.", true],
         ["--grr_username", "GRR username", "admin"],
         ["--grr_password", "GRR password", "admin"],
-        ["--max_file_size", "Maximum size of files to collect (in bytes).", 5368709120]
+        ["--max_file_size", "Maximum size of files to collect (in bytes).", 5368709120],
+        ["--match_mode", "Match mode of the client rule set (ANY or ALL)", null],
+        ["--client_operating_systems", "Comma-separated list of client operating systems to filter hosts on.", null],
```
tomchop, line 34: It would be good to specify the valid options here.

tomchop (changes requested):

tomchop:
Also please fix the mypy issues.

dftimewolf/lib/collectors/grr_hunt.py
sydp: Done

data/recipes/grr_hunt_file.json
sydp: Done.

sydp:
> Also please fix the mypy issues.

Done!

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -54,8 +97,44 @@ def _CreateHunt(
     """
     runner_args = self.grr_api.types.CreateHuntRunnerArgs()
     runner_args.description = self.reason
+
+    if self.match_mode:
+      if self.match_mode.lower() == 'any':
```
tomchop, line 102: We can `lower` the match mode in `HuntSetup()` and assume it's already formatted as expected here.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -25,20 +25,63 @@ class GRRHunt(grr_base.GRRBaseModule, module.BaseModule):  # pylint: disable=abs
...
+
+    Args:
+      match_mode (Optional[str]): match mode of the client rule set (ALL or
+          ANY).
+      client_operating_systems (Optional[str]): a comma separated list of client
+          OS types (win, osx or linux).
+      client_labels (Optional[str]): a comma separated list of client labels.
+    """
+    if match_mode:
+      if match_mode.upper() not in ('ALL', 'ANY'):
```
tomchop, line 61: Super nit: for consistency i'd use either lower or upper throughout this function, but stick to one.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -54,8 +97,44 @@ def _CreateHunt(
...
     runner_args = self.grr_api.types.CreateHuntRunnerArgs()
     runner_args.description = self.reason
+
+    if self.match_mode:
+      if self.match_mode.lower() == 'any':
+        match_mode = runner_args.client_rule_set.MATCH_ANY
+      elif self.match_mode.lower() == 'all':
+        match_mode = runner_args.client_rule_set.MATCH_ALL
+      else:
+        self.ModuleError(f'Unknown match mode {self.match_mode}', critical=True)
```
tomchop, line 107: You're already making sure that match mode is good in HuntSetup(), no? If so, no need to check here.

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -54,8 +97,44 @@ def _CreateHunt(
...
+      for client_label in self.client_labels:
+        label_rule = runner_args.client_rule_set.rules.add()
+
+        label_rule.rule_type = label_rule.LABEL
+        label_rule.label.label_names.append(client_label)
+
+    if self.client_operating_systems:
+      client_operating_systems = set(
+          os for os in self.client_operating_systems
+          if os in ('win', 'osx', 'linux'))
```
tomchop, line 121: Same comment here - I'd say it's safe to assume that `client_operating_systems` has been validated, so no need for th `if` condition here. In fact, you can remove this block completely and just loop over `self.client_operating_systems` in L 123

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -25,20 +25,63 @@ class GRRHunt(grr_base.GRRBaseModule, module.BaseModule):  # pylint: disable=abs
...
+      client_labels (Optional[str]): a comma separated list of client labels.
+    """
+    if match_mode:
+      if match_mode.upper() not in ('ALL', 'ANY'):
+        self.ModuleError(f'Unknown match mode {self.match_mode}', critical=True)
+
+      self.match_mode = match_mode
+
+    if client_operating_systems:
+      normalised_client_operating_systems = list(set(
```
tomchop, line 67: It looks like you're using a `set` in `_CreateHunt`, so you can just skip the `list` here?

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -100,8 +179,11 @@ def SetUp(self,
             grr_username: str,
             grr_password: str,
             max_file_size: str,
-            approvers: Optional[str]=None,
-            verify: bool=True) -> None:
+            approvers: Optional[str] = None,
+            verify: bool = True,
+            match_mode: Optional[str] = None,
+            client_operating_systems: Optional[str] = None,
+            client_labels: Optional[str] = None) -> None:
```
tomchop, line 186: Actually, none of these arguments should be optional (we always provide the full list of kwargs, and ther default values are defined at recipe level). Some cleanup happened but apparently not entirely, heh. Could you remove the `Optional` checks and default values here?

dftimewolf/lib/collectors/grr_hunt.py
```diff
@@ -153,31 +242,34 @@ class GRRHuntFileCollector(GRRHunt):
...
```
tomchop, line 272: Please see my note in the other SetUp function

tomchop (changes requested):

dftimewolf/lib/collectors/grr_hunt.py
sydp: Done.

dftimewolf/lib/collectors/grr_hunt.py
sydp: Switched to lower for consistency.

dftimewolf/lib/collectors/grr_hunt.py
sydp: Yes that's correct.  Removed the check.

dftimewolf/lib/collectors/grr_hunt.py
sydp: Removed.

dftimewolf/lib/collectors/grr_hunt.py
sydp: list() is required to keep mypy happy.

dftimewolf/lib/collectors/grr_hunt.py
sydp: Cleaned up.

dftimewolf/lib/collectors/grr_hunt.py
sydp: Cleaned up.

dftimewolf/lib/collectors/grr_hunt.py
tomchop: This is probably because you type it as a `List[str]` in L44. Have you tried using `Set[str]` instead?

tomchop (changes requested):
One last mypy nit, otherwise this looks good! Thanks a lot :)

sydp:
> One last mypy nit, otherwise this looks good! Thanks a lot :)

Updated!

tomchop (approved):

PR #494, opened 2022-01-07
Consolidate TimesketchExporter and TimesketchExporterThreaded

ramo-j:
Final step of the upgrade of the TimesketchExporter module.

The diff doesn't show this very well, but it's basically `$ git mv -f timesketch_tam.py timesketch.py`


dftimewolf/lib/exporters/timesketch.py
```diff
@@ -152,67 +156,49 @@ def _GetSketchIDFromAttributes(self) -> int:
...
+      streamer.add_file(path)
+      if streamer.response and container.description:
+        streamer.timeline.description = container.description
 
-    api_root = sketch.api.api_root
-    host_url = api_root.partition('api/v1')[0]
-    sketch_url = '{0:s}sketches/{1:d}/'.format(host_url, sketch.id)
+    api_root = self.sketch.api.api_root
+    self.host_url = api_root.partition('api/v1')[0]
+    sketch_url = '{0:s}sketches/{1:d}/'.format(self.host_url, self.sketch.id)
```
tomchop, line 188: Sounds like this bit could be built outside the threaded function, since the results don't depend on the container?

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -152,67 +156,49 @@ def _GetSketchIDFromAttributes(self) -> int:
...
     message = 'Your Timesketch URL is: {0:s}'.format(sketch_url)
-    self.logger.success(message)
-    container = containers.Report(
+    report_container = containers.Report(
         module_name='TimesketchExporter',
         text=message,
         text_format='markdown')
-    self.state.StoreContainer(container)
+    self.state.StoreContainer(report_container)
```
tomchop, line 194: Given that we'll be pushing things to the same sketch, wouldn't it make sense to store just a single report?

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -152,67 +156,49 @@ def _GetSketchIDFromAttributes(self) -> int:
...
 
     if self.wait_for_timelines:
       self.logger.info('Waiting for timelines to finish processing...')
```
tomchop, line 197: It would be great if we could add the `timeline_name` here.

tomchop (changes requested):
Some nits. Btw, I love the direction these ThreadawareModules are taking!

dftimewolf/lib/exporters/timesketch.py
ramo-j: Agreed, moved to PreProcess.

dftimewolf/lib/exporters/timesketch.py
ramo-j: I don't think I agree, the source of this could be multiple GRR collections, multiple turbinia runs, etc.... I think one report per input timeline makes sense.

dftimewolf/lib/exporters/timesketch.py
ramo-j: Now you mention it, I think it makes more sense to move this to PostProcess, since process acts on single timelines within a sketch, the analysers (and waiting for) via `_WaitForTimelines()` acts on the while sketch as one.

ramo-j:
Converting to draft - Currently blocked properly testing these changes by https://github.com/google/timesketch/issues/2103 (Timesketch setup issues after the change from elastic to opensearch.)

dftimewolf/lib/exporters/timesketch.py
ramo-j: Yeah, makes sense. Pushed to `PostProcess`.

tomchop (approved):

PR #497, opened 2022-01-14
Update metawolf's doc

giovannt0:


tomchop (changes requested):
Looks like the linter is complaining a bit

tomchop (approved):

PR #499, opened 2022-01-18
Each run logs in individual files

tomchop:
It started as https://github.com/log2timeline/dftimewolf/pull/498, but this solution was simpler.


ramo-j (approved):
Linter is complaining about some imports no longer being used, but otherwise, LGTM.

ramo-j (approved):

PR #504, opened 2022-01-21
Bugfix in TAM container handling

### LINKED ISSUE #503: Bug in container handling for ThreadAwareModules
Currently, if a module wants to pop input containers, and store the same type of containers after processing, there is a bug that means no containers would be stored.

The relevant functionality is implemented as follows at the moment:

```
        with ThreadPoolExecutor(max_workers=module.GetThreadPoolSize())  executor:
          for container in GetContainers(module.GetThreadOnContainerType()):
            futures.append(submit(module.Process, container))
<snip>
        if not module.KeepThreadedContainersInState():
          self.GetContainers(module.GetThreadOnContainerType(), True)
```

If the module wants to pop it's input containers, they are removed after `Process`. However, if a module (for example, `SCPExporter`) uses the same container type as input and output, then they are added in Process, then removed afterwards.

ramo-j:
Correction & tests for an edge case in container handling for ThreadAwareModules. 

Closes #503

### DIFF
--- a/tests/test_modules/thread_aware_modules.py
+++ b/tests/test_modules/thread_aware_modules.py
@@ -102,3 +102,38 @@ class ThreadAwareConsumerModule(module.ThreadAwareModule):
 
   def PostProcess(self) -> None:
     self.logger.info("ThreadAwareConsumerModule Static Post Process")
+
+class Issue503Module(module.ThreadAwareModule):
+  """This is a module for testing a certain pattern of container handling.
+
+  As described by https://github.com/log2timeline/dftimewolf/issues/503 this
+  module pops containers for input, and uses the same container type as output.
+  """
+  def __init__(self, state, name=None):
+    super(Issue503Module, self).__init__(state, name)
+
+  def SetUp(self): # pylint: disable=arguments-differ
+    """SetUp"""
+    self.logger.info('{0:s} SetUp!'.format(self.name))
+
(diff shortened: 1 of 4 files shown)


tests/test_modules/thread_aware_modules.py
```diff
@@ -102,3 +102,37 @@ def PreProcess(self) -> None:
...
+
+  As described by https://github.com/log2timeline/dftimewolf/issues/503 this
+  module pops containers for input, and uses the same container type as output.
+  """
+  def __init__(self, state, name=None):
+    super(Issue503Module, self).__init__(state, name)
+
+  def SetUp(self): # pylint: disable=arguments-differ
+    """SetUp"""
+    self.logger.info('{0:s} SetUp!'.format(self.name))
```
itsmvd, line 117: Prefer using f-strings over format strings, they're faster and their syntax is easier to read. Same for other format strings in this file.

itsmvd (changes requested):

tests/test_modules/thread_aware_modules.py
itsmvd: As discussed in PM since this is all over this file, lets change this in another PR.

ramo-j: Raised issue #505 for wider discussion on this.

itsmvd (approved):

PR #511, opened 2022-01-28
Move "attributes" attribute from Report to AttributeContainer

### LINKED ISSUE #507: Move "attributes" attribute from from containers.containers.Report to containers.interface.AttributeContainer
This will allow other container types to make use of generic attributes.

jonathan-greig:
Fixes #507


jonathan-greig:
If we're getting rid of the only user of `attributes` (comment in #507) maybe we should rename to `metadata` or something to make it less confusing.

tomchop (approved):

PR #518, opened 2022-02-03
Create AzureLogsCollector module

### LINKED ISSUE #517: Create AzureLogsCollector module
Should save logs in JSON-L format.

jonathan-greig:
Fixes #517

### DIFF
--- a/dftimewolf/lib/collectors/azure_logging.py
+++ b/dftimewolf/lib/collectors/azure_logging.py
@@ -1,0 +1,95 @@
+# -*- coding: utf-8 -*-
+"""Reads logs from an Azure subscription."""
+import json
+import tempfile
+from typing import Optional
+
+from azure import identity as az_identity
+from azure.mgmt import monitor as az_monitor
+from azure.core import exceptions as az_exceptions
+
+from dftimewolf.lib import module
+from dftimewolf.lib.containers import containers
+from dftimewolf.lib.modules import manager as modules_manager
+from dftimewolf.lib.state import DFTimewolfState
+
+
+class AzureLogsCollector(module.BaseModule):
+  """Collector for Azure Activity logs."""
+
+  def __init__(self,
+               state: DFTimewolfState,
+               name: Optional[str]=None,
+               critical: bool=False) -> None:
+    """Initializes a Azure logs collector."""
+    super(AzureLogsCollector, self).__init__(
+        state, name=name, critical=critical)
+    self._filter_expression = ''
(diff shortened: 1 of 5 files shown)


tests/lib/collectors/azure_logging.py
```diff
@@ -0,0 +1,55 @@
...
+
+class AzureLogging(unittest.TestCase):
+  """Tests for the Azure logging collector."""
+
+  def testInitialization(self):
+    """Tests that the collector can be initialized."""
+    test_state = state.DFTimewolfState(config.Config)
+    azure_logging_collector = azure_logging.AzureLogsCollector(test_state)
+    self.assertIsNotNone(azure_logging_collector)
+
```
ramo-j, line 23: I know it's tedious, but we should still test `SetUp` in here to make sure attributes are set correctly.

tests/lib/collectors/azure_logging.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
```
ramo-j, line 1: I'm a big fan of testing error cases, if it's possible to add some in.

dftimewolf/lib/collectors/azure_logging.py
```diff
@@ -0,0 +1,95 @@
...
+
+    credentials = az_identity.DefaultAzureCredential()
+    monitoring_client = az_monitor.MonitorManagementClient(
+        credentials, self._subscription_id)
+    activity_logs_client = monitoring_client.activity_logs
+
+    try:
+      results = activity_logs_client.list(filter=self._filter_expression)
+
+      while True:
```
ramo-j, line 57: What guarantee do we have this won't infinite loop? `StopIteration` *MUST* be thrown at some point?

ramo-j (changes requested):
Just some test cases I'd like to see added mainly, otherwise looks solid to me.

dftimewolf/lib/collectors/azure_logging.py
jonathan-greig: `StopIteration` will be raised when there are no more log entries to be returned for the query (max time range is 90 days). Putting a limit on the loop would limit the number of log entries that could be collected.

tests/lib/collectors/azure_logging.py
jonathan-greig: Done

ramo-j (approved):

PR #527, opened 2022-02-10
GRROsqueryFlowCollector updates (and recipe doc refresh)

sydp:
Updates the attribute data frame container to help distinguish osquery results from other data frames (like CSVs or google sheets).


ramo-j (approved):
LGTM

PR #538, opened 2022-02-17
Filter for meeting_id delimiter

jkppr:
Added a filter that removes the '-' delimiters form the meeting_id when used with the workspace_meet_ts recipe. Additionally I added the info to use the meeting_id without delimiters to the documentations and help as well.

### DIFF
--- a/dftimewolf/lib/collectors/workspace_audit.py
+++ b/dftimewolf/lib/collectors/workspace_audit.py
@@ -126,6 +126,15 @@ class WorkspaceAuditCollector(module.BaseModule):
           for.
       end_time (Optional[str]): End of the time period to return results for.
     """
+
+    # Omit '-' delimiter from the filter_expression (meeting_id) for
+    # the meet application
+    if application_name == 'meet':
+      if '-' in filter_expression:
+        filter_expression = filter_expression.replace('-', '')
+        self.logger.info(
+            'Found \'-\' delimiter in the meeting_id and removed it!')
+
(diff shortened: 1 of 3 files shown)


dftimewolf/lib/collectors/workspace_audit.py
```diff
@@ -126,6 +126,15 @@ def SetUp(self,
           for.
       end_time (Optional[str]): End of the time period to return results for.
     """
+
+    # Omit '-' delimiter from the filter_expression (meeting_id) for
+    # the meet application
+    if application_name == 'meet':
+      if '-' in filter_expression:
```
tomchop, line 133: Maybe combine these in an `and`?

dftimewolf/lib/collectors/workspace_audit.py
```diff
@@ -126,6 +126,15 @@ def SetUp(self,
...
+        filter_expression = filter_expression.replace('-', '')
+        self.logger.info(
+            'Found \'-\' delimiter in the meeting_id and removed it!')
```
tomchop, line 136:
```suggestion
            "Found '-' delimiter in the meeting_id and removed it!")
```

is slightly more readable. You're allowed to infringe the always-use-the-same-quote-type rules when you are including quotes in a string :)

tomchop (changes requested):

jkppr:
Thanks for the input. I adjusted the code accordingly.

tomchop (approved):

PR #539, opened 2022-02-18
Adjusted containers used by GCE Forensics modules

ramo-j:
In the current implementation, `GCEDiskCopy` takes `GCEDisk` containers, pops them, and stores the output containers as the same. This makes it difficult to copy disks to multiple destinations if needed. 

This PR adds a new container type, `GCEDiskEvidence`, which replaces several usages of `GCEDisk`, such as inputs of  `GCEForensicsVM` and `TurbiniaGCPProcessor`.


dftimewolf/lib/processors/gce_forensics_vm.py
```diff
@@ -151,12 +151,15 @@ def Process(self) -> None:
...
-    for d in [d.name for d in self.state.GetContainers(containers.GCEDisk)]:
-      self.logger.info(f'Attaching {d} to {analysis_vm_name}')
-      self.analysis_vm.AttachDisk(compute.GoogleComputeDisk(
-          self.project.project_id,
-          self.project.default_zone,
-          d))
+    disks = self.state.GetContainers(containers.GCEDiskEvidence)
+
+    for d in disks:
+      if d.project == self.project.project_id:
```
itsmvd, line 157: Use a guard clause here to prevent another indentation.

```suggestion
      if d.project != self.project.project_id:
        continue
```

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -62,7 +62,9 @@ def SetUp(self,
     if disk_names:
       for disk in disk_names.strip().split(','):
         if disk:
```
itsmvd, line 64: Guard clause here as well to prevent a third indentation

```suggestion
        if not disk:
          continue
```

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -73,19 +75,21 @@ def SetUp(self,
...
-            containers.GCEDisk(name=container.evidence_disk.name))
+            containers.GCEDiskEvidence(
+                name=container.evidence_disk.name,
+                project=self.project))
 
   # pylint: disable=arguments-renamed
-  def Process(self, disk_container: containers.GCEDisk) -> None:
+  def Process(self, disk_container: containers.GCEDiskEvidence) -> None:
     """Process a GCE Disk with Turbinia."""
     log_file_path = os.path.join(self._output_path, '{0:s}-turbinia.log'.format(
```
itsmvd, line 94: f-string (check elsewhere in the file as well)

PR #540, opened 2022-02-22
GCP Cloud Resource Tree

hkhalifa:
There are 2 recipes added as part of this PR:
- gcp_cloud_resource_tree
- gcp_cloud_resource_tree_offline

Both recipes creates a resource parent/children tree given a GCP cloud resource as an input (currently supported types: gce_instance, gce_disk, gce_image, gce_machine_image, gce_instance_template, gce_snapshot). 

### gcp_cloud_resource_tree: 
- **Summary:** Generates the tree by enumerating all the current resources and builds the tree around the supplied resource.  It also will attempt to fill any gaps identified in the tree through querying the GCP logs
- **Inputs:**
  - project_id : ID of the project where the resource is located
  - resource_id: Resource id
  - resource_name : Resource name
  - resource_type : Resource type (currently supported types: gce_instance, gce_disk, gce_image, gce_machine_image, gce_instance_template, gce_snapshot)
 - **Output**: Table containing information about parent and children resources with a visual tree representation of the relationships. The output will be saved in a file under the temp folder and also dumped on the terminal

### gcp_cloud_resource_tree_offline: 
- **Summary:** Generates the tree using the supplied GCP log file. 
- **Inputs**:
  - project_id : ID of the project where the resource is located
  - resource_id: Resource id
  - resource_name : Resource name
  - resource_type : Resource type (currently supported types: gce_instance, gce_disk, gce_image, gce_machine_image, gce_instance_template, gce_snapshot)
  - paths : Comma-separated paths to GCP log files. Log files should contain log message entries in Json format.
 - **Output**: Table containing information about parent and children resources with a visual tree representation of the relationships. The output will be saved in a file under the temp folder and also dumped on the terminal


data/recipes/gcp_cloud_resource_tree.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "gcp_cloud_resource_tree",
+  "description": "Collects logs from a GCP project and create a resource tree",
```
sa3eed3ed, line 3: add more description to clarify how this work differently that the offline version, i.e. online mode. because both recipes' descriptions are similar to the cmdline user

data/recipes/gcp_cloud_resource_tree_offline.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "gcp_cloud_resource_tree_offline",
+  "description": "Collects logs from a GCP project and create a resource tree",
```
sa3eed3ed, line 3: same as the other recipe, elaborate on what offline mode means

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    Args:
+      project_id: Project id where the resources are located.
+      resource_name: Resource name.
+      resource_type: Resource type (currently supported types: gce_instance,
+        gce_disk, gce_image, gce_machine_image, gce_instance_template,
+        gce_snapshot)
+      mode: operational mode: online or offline
+    """
+    self._project_id = project_id
+    self._resource_name = resource_name
```
sa3eed3ed, line 270: add a quick test to verify resource name is valid else raise module error
you can use the RFC1035 regex in https://github.com/google/cloud-forensics-utils/blob/main/libcloudforensics/providers/gcp/internal/common.py#L41 an example usage in https://github.com/google/cloud-forensics-utils/blob/main/libcloudforensics/providers/gcp/internal/common.py#L99

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+
+    matched_resource = self._FindResource(self._resource_name,
+                                          self._resource_type, None,
+                                          self._project_id)
+
+    if not matched_resource:
+      self.logger.error('Resource not found')
+      return
+
+    resource = matched_resource
```
sa3eed3ed, line 310: Use resource from the beginning, matched_resource variable is redundent

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    elif self._mode == OperatingMode.ONLINE:
+      self.logger.info('Starting module in online mode.')
+      self._GetListOfResources(self._project_id)
+      self._GetResourcesMetaDataFromLogs(self._project_id)
+
+    else:
+      self.PublishMessage(
+          'Invalid operating mode. Supported modes are "online" or "offline"')
+
+    self._BuildResourcesParentRelationships()
```
sa3eed3ed, line 300: move this to SetUp function

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    output = output + dashes + '\n'
+
+    return output
+
+
+class GCPCloudResourceTree(module.BaseModule):
+  """GCP Cloud Resource Tree Creator.
+
+  input: None, takes input from parameters only.
+  output: Temp file with a dump of the resource tree
```
sa3eed3ed, line 228: add type for class attributes, example:
input (str): ...

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+  def SetUp(self, project_id: str, resource_name: str, resource_type: str,
+            mode: str) -> None:
+    """Set up the resource we want to build the tree for.
+
+    Args:
+      project_id: Project id where the resources are located.
+      resource_name: Resource name.
+      resource_type: Resource type (currently supported types: gce_instance,
+        gce_disk, gce_image, gce_machine_image, gce_instance_template,
+        gce_snapshot)
```
sa3eed3ed, line 266: 4 spaces for hanging indent in this codebase

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+      mode: operational mode: online or offline
```
sa3eed3ed, line 267: get rid of the second ":", maybe "mode: The operational mode, which is either online or offline."

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    """For object comparison."""
+    return hash(self.id + self.resource_name)
+
+  @property
+  def resource_name(self) -> str:
+    """Property resource_name."""
+    return self._resource_name
+
+  @resource_name.setter
+  def resource_name(self, value: str) -> None:
```
sa3eed3ed, line 68: it took me a while to understand where does value come from and how it looks like, my understanding is that it should be equal to the resource_name log field (either jsonPayload.resource.name or protoPayload.resourceName), please add docs mentioning this and a link to the docs https://cloud.google.com/compute/docs/logging/migrating-from-activity-logs-to-audit-logs#fields


)

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    """Property resource_name Setter."""
+    if value:
+      values = value.split('/')
+
+      self.name = values[-1]
+
```
sa3eed3ed, line 74: remove whitelines

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+class OperatingMode(Enum):
+  """Enum represent operational mode (Online or Offline)."""
+
+  ONLINE = auto()
+  OFFLINE = auto()
+
+
+class Resource:
+  """A Class that represent a resource (Instance, Disk, Image...etc)."""
```
sa3eed3ed, line 30: move this class to a separate module, something like gcp_cloud_resource_tree_helper.py, dftimewolf tend to keep independent classes and helpers in separate modules to keep the registered modules smaller and easy to understand

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+from libcloudforensics.providers.gcp.internal.log import GoogleCloudLog
+
+from dftimewolf.lib import module
+from dftimewolf.lib.containers import containers
+from dftimewolf.lib.modules import manager as modules_manager
+from dftimewolf.lib.state import DFTimewolfState
+
+
+class OperatingMode(Enum):
+  """Enum represent operational mode (Online or Offline)."""
```
sa3eed3ed, line 23: same as Resource class, move to the helper module

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+  OFFLINE = auto()
+
+
+class Resource:
+  """A Class that represent a resource (Instance, Disk, Image...etc)."""
+
+  has_dynamic_attributes = True  # silences all attribute-errors for Resource
+
+  def __init__(self) -> None:
+    """Initialize the Resource object."""
```
sa3eed3ed, line 35:
```suggestion
    """Initializes the Resource object."""
```

sa3eed3ed, line 35: please add a comment stating that instant variables of this class represents the fields in a log record and link the docs https://cloud.google.com/compute/docs/logging/migrating-from-activity-logs-to-audit-logs

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    self.id = ''
```
sa3eed3ed, line 36: missing type annotation for type str, please add type str here and to all the str arguments below

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+      elif resource_type == 'images':
+        self.type = 'gce_image'
+      elif resource_type == 'machineImages':
+        self.type = 'gce_machine_image'
+      elif resource_type == 'instanceTemplates':
+        self.type = 'gce_instance_template'
+      elif resource_type == 'snapshots':
+        self.type = 'gce_snapshot'
+      else:
+        self.type = resource_type
```
sa3eed3ed, line 89: what is the purpose of matching the default resource_type value from logs to new custom values?
```suggestion
      self.type = values[-2]
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    return hash(self.id + self.resource_name)
+
+  @property
+  def resource_name(self) -> str:
+    """Property resource_name."""
+    return self._resource_name
+
+  @resource_name.setter
+  def resource_name(self, value: str) -> None:
+    """Property resource_name Setter."""
```
sa3eed3ed, line 69: add a comment making it clear that this also set the type

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
sa3eed3ed: Please change all the docstrings in this PR accordingly as per discussed in chat

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+        entry['graph'] = f'{tab*counter}|--{parent_resource.name}'
+        output.insert(0, entry)
+        if parent_resource.parent:
+          parent_resource = parent_resource.parent
+
+    # Add resource entry to the List of dictionaries
+    entry = {}
+    entry['resource_object'] = self
+    entry['graph'] = f'{tab*level}|--{self.name}'
+    output.insert(level, entry)
```
sa3eed3ed, line 168: insert the current entry first to index zero in the list and then the children to have the same order of insertion 

```suggestion
    # Add current resource entry to the List of dictionaries
    entry = {}
    entry['resource_object'] = self
    entry['graph'] = f'{tab*level}|--{self.name}'
    output.insert(0, entry)
    
    # Add the resource parent entries to the List of dictionaries
    counter = level
    parent_resource = self.parent
    if parent_resource:
      while counter > 0:
        counter = counter - 1
        entry: Dict[str, Any] = {}
        entry['resource_object'] = parent_resource
        entry['graph'] = f'{tab*counter}|--{parent_resource.name}'
        output.insert(0, entry)
        if parent_resource.parent:
          parent_resource = parent_resource.parent
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    # Add resource children entries to the List of dictionaries
+    output.extend(self.GenerateChildrenTree(level + 1))
+
+    return output
+
+  def GenerateChildrenTree(self, level: int) -> List[Dict[str, 'Resource']]:
+    """Generate the resource children tree.
+
+    Args:
+      level: The level in the tree to place the children at
```
sa3eed3ed, line 179: dot at the end

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    Returns:
+      List of dictionaries containing a reference to the children resource and
+      their names indented based on their location in the tree
```
sa3eed3ed, line 182: 4 spaces for a hanging indent

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+      state: recipe state.
+      name: The module's runtime name.
+      critical: True if the module is critical, which causes
+          the entire recipe to fail if the module encounters an error.
+    """
+    super(GCPCloudResourceTree, self).__init__(state,
+                                               name=name,
+                                               critical=critical)
+
+    self._project_id = ''
```
sa3eed3ed, line 247: same as in other comment, add type annotation for str

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+        self._resources_dict.values(),
+        key=lambda resource: resource.creation_timestamp)  # type: ignore
+
+    # Building list of timestamp ranges to query
+    time_range = {}
+    if all_active_resources[0].creation_timestamp:
+      time_range['start_timestamp'] = all_active_resources[
+          0].creation_timestamp - timedelta(hours=1)
+      time_range['end_timestamp'] = all_active_resources[
+          0].creation_timestamp + timedelta(hours=1)
```
sa3eed3ed, line 375: 'end_timestamp' key will have a value of  'all_active_resources[
          0].creation_timestamp + timedelta(hours=1)' or  'all_active_resources[
          -1].creation_timestamp + timedelta(hours=1)'
          in other words, do you search in a time range 1 hour before the first stamp and one hour after the first stamp or one hour before the first timestamp and one hour after the last timestamp?

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    """
+    time_ranges: List[Dict[str, datetime]] = []
+
+    all_active_resources = sorted(
+        self._resources_dict.values(),
+        key=lambda resource: resource.creation_timestamp)  # type: ignore
+
+    # Building list of timestamp ranges to query
+    time_range = {}
+    if all_active_resources[0].creation_timestamp:
```
sa3eed3ed, line 371: did you mean to loop on all_active_resources here? or this if condition is intended? this will result in a time_ranges list of length one in all cases, holding single dict with two entries start_timestamp and end_timestamp, from the rest of the function code, I have a feeling this if condition should be changed to a for loop on all_active_resources

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+
+    idx = 1
+    while idx < len(all_active_resources) - 1:
+      time_range = {}
+
+      resource_creation_timestamp = all_active_resources[
+          idx].creation_timestamp
+
+      if not resource_creation_timestamp:
+        return
```
sa3eed3ed, line 386: do you mean here return or continue?
return will exit the method, continue will jump to the next item in the loop

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+        key=lambda resource: resource.creation_timestamp)  # type: ignore
+
+    # Building list of timestamp ranges to query
+    time_range = {}
+    if all_active_resources[0].creation_timestamp:
+      time_range['start_timestamp'] = all_active_resources[
+          0].creation_timestamp - timedelta(hours=1)
+      time_range['end_timestamp'] = all_active_resources[
+          0].creation_timestamp + timedelta(hours=1)
+      time_ranges.append(time_range)
```
sa3eed3ed, line 376: this is the same as the second part in the loop below.
Make idx start from 0 (`idx = 0`) and move this to the loop below, you can make sure this part of the code is entered in the first cycle in the loop by changing condition to:
`if (time_diff.total_seconds() / 60 / 60 / 24) <= 30 and idx>0:`

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+    self._resources_dict.update(self._RetrieveListOfInstances(project_id))
+
+    # Retrieve list of instance templates in a project
+    self._resources_dict.update(
+        self._RetrieveListOfInstanceTemplate(project_id))
+
+    # Retrieve list of machine images in a project
+    self._resources_dict.update(self._RetrieveListOfMachineImages(project_id))
+
+  def _GetResourcesMetaDataFromLogs(self, project_id: str) -> None:
```
sa3eed3ed, line 357: this method does a lot of things, please add high-level description to tasks it does

sa3eed3ed (changes requested):
this is the first round of comments, I am not through the whole PR yet but will stop here and wait your answers

data/recipes/gcp_cloud_resource_tree.json
hkhalifa: Done

data/recipes/gcp_cloud_resource_tree_offline.json
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: resource_name is a property of Resource. "value" is the value you set it to for example resource.resource_name = "anything".

I added a comment with examples of how value is expected.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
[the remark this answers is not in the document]
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: it doesn't necessarily represent the fields in a log record. It could be a disk resource for example obtained using compute api. I already have """A Class that represents a resource (Instance, Disk, Image...etc).""" cause this is a custom class does not necessarily map 1:1 to what's in the proto/log record.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: The type will be used in the command line input by the user where they supply the resource name and resource type. I created those to make the type values more user friendly. so instead of supplying "dm2 machineImages" in the argument, the user can supply "dm2 gcp_machine_image". The camelcase is not intuitive and having the plural of the type doesn't make sense for the user.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+        'Deleter IP Addr', 'Tree')
+    output = output + dashes + '\n'
+
+    result = self.GenerateTree()
+    for i in result:
+      resource: Optional[Resource] = i.get('resource_object')
+      if resource:
+        output = output + \
+            ('{:<20s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<100s}\n'.format( resource.id if ('-' not in resource.id) else "" , resource.type, resource.creation_timestamp.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S") if resource.creation_timestamp else "",
+            resource.created_by, resource.creator_ip_address,  resource.deletion_timestamp.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S") if resource.deletion_timestamp else "", resource.deleted_by, resource.deleter_ip_address,  i.get('graph')))
```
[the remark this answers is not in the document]
hkhalifa: These are all initialised with empty string. If they are not set, then it's fine.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Not sure what you want to achieve with that change. It seems to be the same for me.  I just add things in order they are printed: parent resource, resource in question and then children resource for things to be more readable.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+      time_range = {}
+
+      resource_creation_timestamp = all_active_resources[
+          idx].creation_timestamp
+
+      if not resource_creation_timestamp:
+        return
+
+      index_of_last_timestamp = len(time_ranges) - 1
+      end_timestamp_of_last_time_range = time_ranges[index_of_last_timestamp][
```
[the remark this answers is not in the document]
hkhalifa: This was my lame attempt to make things more readable when reading the logical, but -1 will be clear as well.

ramo-j (dismissed):
Drive by review - Please make sure you add tests for this new module.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: I refactored this function

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+          'end_timestamp']
+
+      time_diff = resource_creation_timestamp - end_timestamp_of_last_time_range
+      # To optimize calls to retrieve logs, we attempt to group resources that
+      # have a creation timestamp within 30 days.
+      if (time_diff.total_seconds() / 60 / 60 / 24) <= 30:
+        time_range['start_timestamp'] = time_ranges[index_of_last_timestamp][
```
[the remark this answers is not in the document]
hkhalifa: Refactored this function

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done. I was following https://peps.python.org/pep-0257/ : "The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: There is no need for the test. I am not creating new resources. Few lines below, I search for the resource and if not found I display an error message.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1199 @@
...
+      elif resource_type == 'machineImages':
+        self.type = 'gce_machine_image'
+      elif resource_type == 'instanceTemplates':
+        self.type = 'gce_instance_template'
+      elif resource_type == 'snapshots':
+        self.type = 'gce_snapshot'
+      else:
+        self.type = resource_type
+
+      if '/zones/' in value:
```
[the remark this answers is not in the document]
hkhalifa: Didn't consider regional disk. Will keep this comment open till I test and add it.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
sa3eed3ed: got it

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,988 @@
...
+    self.resources_dict: Dict[str, Resource] = {}
+
+  # pylint: disable=arguments-differ
+  def SetUp(self, project_id: str, resource_name: str, resource_type: str,
+            mode: str) -> None:
+    """Sets up the resource we want to build the tree for.
+
+    Args:
+      project_id: Project id where the resources are located.
+      resource_name: Resource name.
```
sa3eed3ed, line 61: is there a reason why not to have resource id as a parameter as well and make sure the user provide either resource id or resource name is available? sometimes there might be two disks or instances with the same name because one was deleted and then the next was created, in this case the id is the way to differentiate between these two instances, also when automation take-care of starting and shutting down resources (ex: managed groups) this case will happen often

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+import json
+import tempfile
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
+from typing import Dict, List, Optional, Any
+
+from libcloudforensics.providers.gcp.internal.common import CreateService
+from libcloudforensics.providers.gcp.internal.log import GoogleCloudLog
```
sa3eed3ed, line 12: Please don't import names from modules
Use import x for importing packages and modules.
Use from x import y where x is the package prefix and y is the module name

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+from dftimewolf.lib import module
+from dftimewolf.lib.containers import containers
+from dftimewolf.lib.modules import manager as modules_manager
+from dftimewolf.lib.state import DFTimewolfState
```
sa3eed3ed, line 17: same here as in the last comment

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+from dftimewolf.lib.processors.gcp_cloud_resource_tree_helper import OperatingMode, Resource  # pylint: disable=line-too-long
```
sa3eed3ed, line 19: same here as in the last comment

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+class GCPCloudResourceTree(module.BaseModule):
+  """GCP Cloud Resource Tree Creator.
+
+  Attributes:
+    project_id (str): Id of the project where the resource is located.
+    resource_name (str): Name of the resource to build the tree for.
+    resource_type (str): Resource type.
+    resources_dict (Dict[str, Resource]): Dictionary of resources
+    period_covered_by_retrieved_logs (Dict[str, datetime]): Dictionary of the time
```
sa3eed3ed, line 30: add mode to the list of attributes

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
sa3eed3ed: you can also use import x from y as z
to use a short alias

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+      critical: True if the module is critical, which causes
+          the entire recipe to fail if the module encounters an error.
+    """
+    super(GCPCloudResourceTree, self).__init__(state,
+                                               name=name,
+                                               critical=critical)
+
+    self.project_id: str = ''
+    self.resource_name: str = ''
+    self.resource_type: str = ''
```
sa3eed3ed, line 53: are all these required parameters, if yes why a default value of empty string? making them required will guarantee that the user pass a value otherwise an empty value will not fail in setup which should be the case

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+    if not file_container.path:
+      self.logger.error('File container path is null or empty')
+
+    with open(file_container.path, 'r') as input_file:
+      log_messages = []
+      file_content = input_file.readline()
+
+      if not file_content:
+        self.logger.error(f'The supplied file {file_container.path} is empty')
+        return
```
sa3eed3ed, line 394: raise a critical module error before return, there is no need to continue here

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+    if resource_id:
+      query_filter = query_filter + \
+          f' AND (resource.labels.instance_id="{resource_id}" OR resource.labels.image_id="{resource_id}")'
+
+    log_messages: List[Dict[str,
+                            Any]] = gcl.ExecuteQuery(qfilter=[query_filter])
+
+    return log_messages
+
+  def _ParseLogMessagesFromFileContainer(
```
sa3eed3ed, line 373: just out of curiosity, is the the expected file format one log entry per line? is it the default format for downloaded gcp logs?

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+      while file_content:
+        log_messages.append(json.loads(file_content))
+        file_content = input_file.readline()
+
+      self._ParseLogMessages(log_messages)
+
+  def _ParseLogMessages(self, log_messages: List[Dict[str, Any]]) -> None:
+    """Parses supplied GCP Cloud log messages and fills self._resource_dict with
+    the result.
```
sa3eed3ed, line 404: function description in one line. you can follow on that with a more detailed description in a separate paragraph (with one blank line in-between)

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
sa3eed3ed: Om I see now, its a json file

data/recipes/gcp_cloud_resource_tree_offline.json
```diff
@@ -0,0 +1,27 @@
...
+          "resource_name": "@resource_name",
+          "resource_type": "@resource_type",
+          "mode": "offline"
+      }
+  }],
+  "args": [
+      ["project_id", "ID of the project where the resource is located", null],
+      ["resource_name","Resource name", null],
+      ["resource_type","Resource type (currently supported types: gce_instance, gce_disk, gce_image, gce_machine_image, gce_instance_template, gce_snapshot)", null],
+      ["paths", "Comma-separated paths to GCP log files", null]
```
sa3eed3ed, line 25: add the expected file format, ex: json, to the description on the argument

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+        if not resource:
+          resource = Resource()
+          resource.id = response.get('targetId')
+
+        # compute.disks.createSnapshot is a special case where the
+        # "resourceName" is just the name and not the full name with the
+        # project, zone, type and name
+        if log_message_type.startswith('compute.disks.createSnapshot'):
+          resource = Resource()
+          source_disk_id = response.get('targetId')
```
sa3eed3ed, line 446: move `source_disk_id`, `resource.project_id` and `resource.zone` value assignment to the else branch of the next if condition, these will be overridden with the matched_resource in case it is not empty, thus no need to assign them before the if

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+          self.resources_dict[resource.id] = resource
+
+  def _ParseInsertLogMessage(self, resource: Resource, request: Dict[str, Any],
+                             response: Dict[str, Any]) -> None:
+    """Parses a GCP log message where the operation is insert or create.
+
+    Args:
+      resource: Resource object to update with parsed information
+      request: Request portion of the log message
+      response: Response portion of the log message
```
sa3eed3ed, line 503: add dots at the end of each sentence

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+    """
+    resource.creation_timestamp = response.get('insertTime')
+    resource.created_by = response.get('user', '')
+
+    if 'sourceDisk' in request:
+      if not resource.parent:
+        resource.parent = Resource()
+        resource.parent.resource_name = request.get('sourceDisk', '')
```
sa3eed3ed, line 511:
```suggestion
    if 'sourceDisk' in request and not resource.parent:
      resource.parent = Resource()
      resource.parent.resource_name = request.get('sourceDisk', '')
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+    elif 'sourceInstance' in request:
+      if not resource.parent:
+        resource.parent = Resource()
+        resource.parent.resource_name = request.get('sourceInstance', '')
+
+    # When creating a disk image from another disk
+    elif 'sourceSnapshot' in request:
+      if not resource.parent:
+        resource.parent = Resource()
+        resource.parent.resource_name = request.get('sourceSnapshot', '')
```
sa3eed3ed, line 523: check the resource.parent once instead of three times
```suggestion
    if not resource.parent:
      if 'sourceDisk' in request:
        resource.parent = Resource()
        resource.parent.resource_name = request.get('sourceDisk', '')

      # When creating machine image from an instance
      elif 'sourceInstance' in request:
        resource.parent = Resource()
        resource.parent.resource_name = request.get('sourceInstance', '')

      # When creating a disk image from another disk
      elif 'sourceSnapshot' in request:
        resource.parent = Resource()
        resource.parent.resource_name = request.get('sourceSnapshot', '')
```

sa3eed3ed: what if there is no parent and none of these is in the request, should the function just continue or do you need to stop?

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+    if request.get('disks'):
+      for disk in request.get('disks', {}):
+        disk_resource = Resource()
+        disk_resource.project_id = resource.project_id
+        disk_resource.created_by = resource.created_by
+        disk_resource.creator_ip_address = resource.creator_ip_address
+        disk_resource.creator_useragent = resource.creator_useragent
+        disk_resource.zone = resource.zone
+        disk_resource.name = disk.get('deviceName')
+        disk_resource.type = 'gce_disk'
```
sa3eed3ed, line 537: consider sending values via the constructor while initializing the method, as it is easier to read
```suggestion
        disk_resource = Resource(
            name=disk.get('deviceName'),
            type='gce_disk',
            project_id=resource.project_id,
            zone=resource.zone,
            created_by=resource.created_by,
            creator_ip_address=resource.creator_ip_address,
            creator_useragent=resource.creator_useragent
        )
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+          resource.disks.append(matched_resource)
+          continue
+        # pylint: enable=line-too-long
+
+        initialize_params = disk.get('initializeParams')
+        if initialize_params:
+          # When instance is created using an existing disk image, machine image
+          # or template
+          if 'sourceImage' in initialize_params:
+            if not disk_resource.parent:
```
sa3eed3ed, line 567: make if not disk_resource.parent: the outer if condition and below it the three cases to not do any further check if no disk_resource.parent

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+      else:
+        self.PublishMessage(f'Type {log_message_type} not supported')
+        resource = None
+
+      if resource:
+        if resource.id:
+          self.resources_dict[resource.id] = resource
+
+  def _ParseInsertLogMessage(self, resource: Resource, request: Dict[str, Any],
```
sa3eed3ed, line 496: please add some description that this function changes the passed Resource object in-place with the information parsed from request and response arguments

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+        return
+
+      while file_content:
+        log_messages.append(json.loads(file_content))
+        file_content = input_file.readline()
+
+      self._ParseLogMessages(log_messages)
+
+  def _ParseLogMessages(self, log_messages: List[Dict[str, Any]]) -> None:
+    """Parses supplied GCP Cloud log messages and fills self._resource_dict with
```
sa3eed3ed, line 403: here you mention that it fills `self._resource_dict` while it updates `self.resource_dict` and not `self._resource_dict`

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+          gce_snapshot)
+      mode: operational mode (online or offline)
+    """
+    self.project_id = project_id
+    self.resource_name = resource_name
+    self.resource_type = resource_type
+    if 'offline' in mode:
+      self.mode = OperatingMode.OFFLINE
+    elif 'online' in mode:
+      self.mode = OperatingMode.ONLINE
```
sa3eed3ed, line 77: if mode is not offline or online you should raise an error and quite here and not in process, i.e. fail early 
dftimewolf try to do all the quick verifications in the setup method, setup methods from all modules in a recipe are run first, the process method will only be run after all setup methods of all modules and all process methods of the prior modules are run which can lead to waiting for 5 mins and then getting a message "invalid mode"

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+      self._ParseLogMessages(log_messages)
+
+      matched_resource = self._FindResource(resource.name, resource.type)
+      if matched_resource:
+        if matched_resource.deletion_timestamp and matched_resource.creation_timestamp:  # pylint: disable=line-too-long
+          return matched_resource
+
+    return None
+
+  def _RetrieveListOfDisks(self, project_id: str) -> Dict[str, Resource]:
```
sa3eed3ed, line 660: why don't you use this https://github.com/google/cloud-forensics-utils/blob/719093b4a229e5e97c30d93faabb1ccf3b6ee422/libcloudforensics/providers/gcp/internal/compute.py#L108

this adds code that wraps gcp APIs while we have libcloudforensics for that and it is used in dftimewolf for any cloud api operation

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+            resource.parent = Resource()
+            resource.parent.resource_name = disk.get('sourceImage')
+            resource.parent.id = disk.get('sourceImageId')
+
+          result[resource.id] = resource
+
+      # pylint: disable=no-member
+      request = compute_api_client.disks().aggregatedList_next(
+          previous_request=request, previous_response=response)
+
```
sa3eed3ed, line 705: calling `ListDisks` https://github.com/google/cloud-forensics-utils/blob/719093b4a229e5e97c30d93faabb1ccf3b6ee422/libcloudforensics/providers/gcp/internal/compute.py#L242 
will do all that and it will return a dict of Dict[disk_name[str]: disk[GoogleComputeDisk]]
you can then do disk.GetOperation() and get the full api object of that disk or you can call .keys() on the dict and get names of all disks

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+          result[resource.id] = resource
+
+      # pylint: disable=no-member
+      request = compute_api_client.snapshots().list_next(
+          previous_request=request, previous_response=response)
+
+    return result
+
+  def _RetrieveListOfInstances(self, project_id: str) -> Dict[str, Resource]:
```
sa3eed3ed, line 791: this can be replaced by function in https://github.com/google/cloud-forensics-utils/blob/main/libcloudforensics/providers/gcp/internal/compute.py

hkhalifa: Ya i know. At the time I wrote this part of the code some of the required fields were only in compute engine beta api and not in the v1 used by libcloudforensics. So i had to do it myself. Will try libcloudforensics and check if v1 has the fields i want.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+          result[resource.id] = resource
+
+      # pylint: disable=no-member
+      request = compute_api_client.images().list_next(
+          previous_request=request, previous_response=response)
+
+    return result
+
+  def _RetrieveListOfSnapshots(self, project_id: str) -> Dict[str, Resource]:
```
sa3eed3ed, line 752: same as _RetrieveListOfDiskImages

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+          result[resource.id] = resource
+
+      # pylint: disable=no-member
+      request = compute_api_client.instances().aggregatedList_next(
+          previous_request=request, previous_response=response)
+
+    return result
+
+  def _RetrieveListOfInstanceTemplates(self,
```
sa3eed3ed, line 869: same as _RetrieveListOfDiskImages

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+          time_diff = resource.creation_timestamp - time_ranges[-1][
+              'end_timestamp']
+          # To optimize calls to retrieve logs, we attempt to group resources
+          # that have a creation timestamp within 30 days.
+          if (time_diff.total_seconds() / 60 / 60 / 24) <= 30:
+            time_range['start_timestamp'] = time_ranges[-1]['start_timestamp']
+            time_range[
+                'end_timestamp'] = resource.creation_timestamp + timedelta(
+                    hours=1)
+            time_ranges[-1] = time_range
```
sa3eed3ed, line 200:
```suggestion
            time_ranges[-1][
                'end_timestamp'] = resource.creation_timestamp + timedelta(
                    hours=1)
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: libcloudforensics doesn't have the ability to list snapshots

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: libcloudforensics doesn't have the ability to list Instance Templates

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+
+          result[resource.id] = resource
+
+      # pylint: disable=no-member
+      request = compute_api_client.instanceTemplates().list_next(
+          previous_request=request, previous_response=response)
+
+    return result
+
+  def _RetrieveListOfMachineImages(self,
```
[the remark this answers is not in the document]
hkhalifa: libcloudforensics doesn't have the ability to list Machine Images

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+      project_id: Project id from which we are going to obtain the logs.
+      start_timestamp: Retrieve logs starting at this timestamp.
+      end_timestamp: Retrieve logs ending at this timestamp.
+      resource_id: Optional resource id
+
+    Return:
+      List of log messages
+    """
+    if not project_id or not start_timestamp or not end_timestamp:
+      return []
```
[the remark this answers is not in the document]
hkhalifa: Removed

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Added the ability to supply resource_id

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1002 @@
...
+        'start') or not self.period_covered_by_retrieved_logs.get('end'):
+      self.period_covered_by_retrieved_logs['start'] = start_timestamp
+      self.period_covered_by_retrieved_logs['end'] = end_timestamp
+    else:
+      # If the required time range is within the time range already retrieved
+      # before return an empty list
+      if start_timestamp >= self.period_covered_by_retrieved_logs[
+          'start'] and end_timestamp <= self.period_covered_by_retrieved_logs[
+              'end']:
+        return []
```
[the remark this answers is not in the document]
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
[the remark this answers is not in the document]
hkhalifa: Ya i know. At the time I wrote this part of the code some of the required fields were only in compute engine beta api and not in the v1 used by libcloudforensics. So i had to do it myself. Will try libcloudforensics and check if v1 has the fields i want. Also libcloudforensics has bug that will affect the outcome (https://github.com/google/cloud-forensics-utils/pull/440)

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

data/recipes/gcp_cloud_resource_tree_offline.json
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: I removed "if not resource.parent:" It was mainly there to initialise the resource object if it's null. I just initialise it anyway. And yes if no parent I still want to continue to parse disks.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Same as the one before removed the check and initialised anyway

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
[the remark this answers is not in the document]
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: I use the same style across, cause most of the time I keep adding to the Resource object as I parse data. So yes i can change that, but don't see the value and keeping it as is will be consistent across the code.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1029 @@
...
+    Args:
+      project_id: Project id where the resource are located.
+      zone: zone where the resource are located.
+      resource_id: Resource id.
+      resource_name: Resource name.
+      resource_type: Resource type (currently supported types: gce_instance,
+          gce_disk, gce_image, gce_machine_image, gce_instance_template,
+          gce_snapshot)
+      mode: operational mode (online or offline)
+    """
```
sa3eed3ed, line 78: Making resource_id and resource_name optional as these can be None
```suggestion
  def SetUp(self,
                     project_id: str, zone: str,
                     resource_type: str,
                     mode: str,
                     resource_id: Optional[str] = None,
                     resource_name: Optional[str] = None) -> None:
    """Sets up the resource we want to build the tree for.

    Args:
      project_id: Project id where the resource are located.
      zone: zone where the resource are located.
      resource_type: Resource type (currently supported types: gce_instance,
          gce_disk, gce_image, gce_machine_image, gce_instance_template,
          gce_snapshot)
      mode: operational mode (online or offline).
      resource_id: Resource id.
      resource_name: Resource name.
    """
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+    self.zone = zone
+    self.resource_id = resource_id
+    self.resource_name = resource_name
+    self.resource_type = resource_type
+    if 'offline' in mode:
+      self.mode = gcp_crt_helper.OperatingMode.OFFLINE
+    elif 'online' in mode:
+      self.mode = gcp_crt_helper.OperatingMode.ONLINE
+    else:
+      self.ModuleError("Operational mode not set.")
```
sa3eed3ed, line 97: quit if operational mode is not set, because code will do nothing after
```suggestion
      self.ModuleError("Operational mode not set.", critical=True)
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      return
+
+    # Save resource tree to temp file
+    output_file = tempfile.NamedTemporaryFile(mode='w',
+                                              delete=False,
+                                              encoding='utf-8',
+                                              suffix='.txt')
+    output_path = output_file.name
+
+    self.PublishMessage(f'Saving resource tree to {output_path}')
```
sa3eed3ed, line 147: below you check if there is parent and if this is the case you str(parent), why do you do that? as you override __str__ and call generate tree as part of it, which loops on all parents and children then add them to the graph, won't it be enough if you just call str(parent) in all cases

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+    # resource. This is filled during the parsing of log messages in
+    # _ParesLogMessages() and/or _GetListOfResources
+    if resource.parent and resource.parent.name and resource.parent.type:
+      if resource.parent.id:
+        parent_resource = self.resources_dict.get(resource.parent.id)
+        # If the parent resource is deleted or it's one of the stock disk images
+        # (for ex Debian), we will have the parent id but it's not in the list
+        # of resources we parsed.
+        if not parent_resource:
+          parent_resource = resource.parent
```
sa3eed3ed, line 271:
```suggestion
      if resource.parent.id:
        # If the parent resource is deleted or it's one of the stock disk images
        # (for ex Debian), we will have the parent id but it's not in the list
        # of resources we parsed.
        parent_resource = self.resources_dict.get(resource.parent.id,
            resource.parent)
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+    Return:
+      Resource object that match the name and type or None if a matching
+      resource is not found
+    """
+    # Search for the resource with the same name and type in the parsed logs.
+    for resource in self.resources_dict.values():
+
+      if resource.name == resource_name and resource.type == resource_type:
+        # Filter list by zone if it is supplied
+        if zone is not None and resource.zone != zone:
```
sa3eed3ed, line 323:
```suggestion
        if zone and resource.zone != zone:
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+          continue
+        # Check is project id match if project id was supplied as a filter
+        # criteria
+        if project_id is not None and resource.project_id != project_id:
```
sa3eed3ed, line 327:
```suggestion
        if project_id and resource.project_id != project_id:
```

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      self._ParseLogMessages(log_messages)
+
+      matched_resource = self._FindResource(resource.name, resource.type)
+      if matched_resource:
+        if matched_resource.deletion_timestamp and matched_resource.creation_timestamp:  # pylint: disable=line-too-long
+          return matched_resource
+
+    return None
+
+  def _RetrieveListOfDisks(
```
sa3eed3ed, line 688: use ListDisks or project.Disks from libcloudforensics, this will give you all zonal disks, regional disks can be obtained via ListRegionDisks and they have a region field but not a zone

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      project_id: Project Id to retrieve the list of disk images for
+
+    Return:
+      Dict of disk images
+    """
+    result: Dict[str, gcp_crt_helper.Resource] = {}
+
+    # Using beta version of the API because v1 did not have important
+    # information when creating this script
+    compute_api_client = gcp_common.CreateService('compute', 'beta')
```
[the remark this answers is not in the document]
sa3eed3ed: please open an issue to move this to libcloudforensics later

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      project_id: Project Id to retrieve the list of snapshots for
+
+    Returns:
+      Dict of snapshots
+    """
+    result: Dict[str, gcp_crt_helper.Resource] = {}
+
+    # Using beta version of the API because v1 did not have important
+    # information when creating this script
+    compute_api_client = gcp_common.CreateService('compute', 'beta')
```
sa3eed3ed, line 796: same as images

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      project_id: Project Id to retrieve the list of instances for
+
+    Returns:
+      Dict of instances
+    """
+    result: Dict[str, gcp_crt_helper.Resource] = {}
+
+    # Using beta version of the API because v1 did not have important
+    # information when creating this script
+    compute_api_client = gcp_common.CreateService('compute', 'beta')
```
sa3eed3ed, line 836: use ListInstances in lcf, it will give you instances keyed by ID after https://github.com/google/cloud-forensics-utils/pull/440

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      project_id: Project Id to retrieve the list of instance templates for
+
+    Returns:
+      Dict of instance templates
+    """
+    result: Dict[str, gcp_crt_helper.Resource] = {}
+
+    # Using beta version of the API because v1 did not have important
+    # information when creating this script
+    compute_api_client = gcp_common.CreateService('compute', 'beta')
```
sa3eed3ed, line 915: same as images and snapshots

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+      project_id: Project Id to retrieve the list of machine images for
+
+    Returns:
+      Dict of machine images
+    """
+    result: Dict[str, gcp_crt_helper.Resource] = {}
+
+    # Using beta version of the API because v1 did not have important
+    # information when creating this script
+    compute_api_client = gcp_common.CreateService('compute', 'beta')
```
sa3eed3ed, line 1002: same like snapshots

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1034 @@
...
+    if not file_container.path:
+      self.logger.error('File container path is null or empty')
+      return
+
+    with open(file_container.path, 'r') as input_file:
+      log_messages = []
+      file_content = input_file.readline()
+
+      if not file_content:
+        self.logger.error(f'The supplied file {file_container.path} is empty')
```
sa3eed3ed, line 421:
```suggestion
        self.logger.warning(f'The supplied file {file_container.path} is empty')
```

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
```diff
@@ -0,0 +1,303 @@
...
+    parent_resource = self.parent
+    if parent_resource:
+      while counter > 0:
+        counter = counter - 1
+        entry: Dict[str, Any] = {}
+        entry['resource_object'] = parent_resource
+        entry['graph'] = f'{tab*counter}|--{parent_resource.name}'
+        output.insert(0, entry)
+        if parent_resource.parent:
+          parent_resource = parent_resource.parent
```
sa3eed3ed, line 216: you start here from self.parent, create entry and then insert (0,entry), so you insert the entries in the following order: parent, grandparent, grandgrandparent, ...

then after the loop finishes you create entry for self and push it to the end of output with insert(level,entry). 

Why don't you move adding the resource entry before adding parents and insert it to (0,entry)?

this way the insertion happens in order and less confusing for reader

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
```diff
@@ -0,0 +1,303 @@
...
+      if resource:
+        output = output + \
+            ('{:<20s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<100s}\n'.format( resource.id if ('-' not in resource.id) else "" , resource.type, resource.creation_timestamp.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S") if resource.creation_timestamp else "",
+            resource.created_by, resource.creator_ip_address,  resource.deletion_timestamp.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S") if resource.deletion_timestamp else "", resource.deleted_by, resource.deleter_ip_address,  i.get('graph')))
+    output = output + dashes + '\n'
+
+    return output
+
+
+class ResourceEncoder(json.JSONEncoder):
```
sa3eed3ed, line 278: is this just for testing or is it used anywhere else? if it is only for testing move it to the test class

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Maybe later on when other methods like ListSnapshots, ListMachineImages, ListDiskImages and ListInstanceTemplate are implemented. Doing some (ListDisks and ListInstances) one way and the other in another way, might be confusing.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: I answered this comment before

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
hkhalifa: I already replied to the same comment here https://github.com/log2timeline/dftimewolf/pull/540#discussion_r917949500 

TL;DR, doesn't make a difference. I add things in the same order they are displayed. Parent resources > Resource in question > Child Resources. Feels more readable to me. Also I need to determine the 'level' the resource is at which I use to determine the indentation. I don't know how many parents are there before hand.

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
hkhalifa: It's an encoder for the Resource object. It not test specific, but currently it's only used in tests, so I will mode it there.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: No, when you run str(reource) it only gets you the tree related to this resource, so for example you won't see the sibling of the resource. I check if the resource has a parent and if so I display the tree of the parent so sibling will be visibile which i think might be helpfull for the analyst. If the reource doesn't have a parent i just display the tree for the resource.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: @sa3eed3ed please check if I should mark this as resolved.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: @sa3eed3ed please check if I should mark this as resolved. Or do you want me to still do the change.

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: @sa3eed3ed please check if I should mark this as resolved.

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
```diff
@@ -0,0 +1,303 @@
...
+      self._deletion_timestamp = value
+    else:
+      self._deletion_timestamp = parser.parse(value)
+
+  def IsDeleted(self) -> bool:
+    """Checks if resource is deleted."""
+    if self.deleted or self.deletion_timestamp or not self.creation_timestamp:
+      return True
+
+    return False
```
sa3eed3ed, line 185: return the condition directly
```suggestion
    return self.deleted or self.deletion_timestamp or not self.creation_timestamp
```

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
```diff
@@ -0,0 +1,303 @@
...
+class Resource():
+  # pylint: disable=line-too-long
+  """A Class that represents a resource (Instance, Disk, Image...etc).
+
+  Attributes:
+    id (str): Id of the resource.
+    name (str): Name of the resource.
+    type (str): Resource type.
+    project_id (str): Id of the project where the resource is located.
+    zone (str): Zone/region where the resource is located.
```
sa3eed3ed, line 25: Optional: name this location

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
```diff
@@ -0,0 +1,303 @@
...
+        'Creator IP Addr', 'Deletion Timestamp', 'Deleted By',
+        'Deleter IP Addr', 'Tree')
+    output = output + dashes + '\n'
+
+    result = self.GenerateTree()
+    for i in result:
+      resource: Optional[Resource] = i.get('resource_object')
+      if resource:
+        output = output + \
+            ('{:<20s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<100s}\n'.format( resource.id if ('-' not in resource.id) else "" , resource.type, resource.creation_timestamp.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S") if resource.creation_timestamp else "",
```
sa3eed3ed, line 271: as you use this string more than once, you can make it a local variable and use it here and in line 260, it will also help shorten the line
'{:<20s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<20s}|{:<25s}|{:<18s}|{:<100s}\n'

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
sa3eed3ed: if you think it can be used later feel free to leave it, this is Optional

sa3eed3ed (approved):
LGTM and small nits, please fix the broken flows

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
```diff
@@ -0,0 +1,1032 @@
...
+      project_id: Project Id to retrieve the list of disks for
+
+    Returns:
+      Dict of disks
+    """
+    result: Dict[str, gcp_crt_helper.Resource] = {}
+
+    # Using beta version of the API because v1 did not have important
+    # information when creating this script
+    compute_api_client = gcp_common.CreateService('compute', 'beta')
```
sa3eed3ed, line 700: you create a new client object and do full auth handshake with every call to CreateService, you can cache the compute client once in the setup method (if no creds this will fails which is another reason to do this in setup and avoid failing late), for example either:
- make it instance variable, cache client in setup 
`self.compute_api_client = gcp_common.CreateService('compute', 'v1')`
- or use (it defaults to v1):
sleft.compute_api_client = gcp_common.GoogleCloudComputeClient(project_id).GceApi()

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
hkhalifa: Mypy complains when I do that. Also I think it's more readable in the current state.

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: I will create one instance in `_GetListOfResources` and pass it to each of the resource. As for the check, I have a preflight check in the recipe to check if the user has access to the project.

dftimewolf/lib/processors/gcp_cloud_resource_tree_helper.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
hkhalifa: Done

dftimewolf/lib/processors/gcp_cloud_resource_tree.py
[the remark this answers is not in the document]
hkhalifa: Will open an issue.

PR #544, opened 2022-03-02
GRRFlowCollector - Error handling improvements

ramo-j:
Updated GRRFlowCollector to raise an error when no flows are found. This can occur if the user specifies a child flow rather than a parent flow, which is no longer supported.

In SetUp, each flow that is not found will be noted as a warning:

```
[2022-03-02 02:51:37,686] [GRRFlowCollector    ] WARNING  The following flows were not found: F:23456, F:34567
[2022-03-02 02:51:37,687] [GRRFlowCollector    ] WARNING  Did you specify a child flow instead of a parent?
```

 In PreProcess, if no flows are found at all, an error is raised. SetUp doesn't error in the event the containers are coming from a previous module.

```
[2022-03-02 02:55:50,353] [GRRFlowCollector    ] CRITICAL No flows found for collection.
```

In Process, if a flow has no data collected:

```
[2022-03-02 03:08:47,585] [GRRFlowCollector    ] WARNING  [MainThread] No flow data collected for C.0000000000000001:F:12345
```

### DIFF
--- a/dftimewolf/lib/collectors/grr_hosts.py
+++ b/dftimewolf/lib/collectors/grr_hosts.py
@@ -882,6 +882,7 @@ class GRRFlowCollector(GRRFlow):
         skip_offline_clients=skip_offline_clients)
 
     flows = flow_ids.strip().split(',')
+    found_flows = []
 
     # For each host specified, list their flows
     for item in hostnames.strip().split(','):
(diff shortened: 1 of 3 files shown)


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -893,6 +894,13 @@ def SetUp(self,
         for f in flows:
           if f in client_flows:
             self.state.StoreContainer(containers.GrrFlow(host, f))
+            found_flows.append(f)
+
+    missing_flows = sorted([f for f in flows if f not in found_flows])
+    if len(missing_flows) != 0:
```
itsmvd, line 900:
```suggestion
    if missing_flows:
```

This is debatable, but I prefer it like this. Seems more pythonic.

itsmvd (approved):

PR #545, opened 2022-03-02
TurbiniaGCP - Ignore disks in other projects

ramo-j:
Currently the TurbiniaGCPProcessor module doesn't consider the project in the GCEDiskEvidence container, meaning it may attempt to process disks in other projects. This change corrects that behaviour.


dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -92,6 +92,9 @@ def PreProcess(self) -> None:
   # pylint: disable=arguments-renamed
   def Process(self, disk_container: containers.GCEDiskEvidence) -> None:
     """Process a GCE Disk with Turbinia."""
+    if disk_container.project != self.project:
+      return
```
aarontp, line 96: Should we add a logging statement here to say that we're not processing this because it's not in the target project?  It might be confusing to see it start and end without knowing why.

ramo-j: Good call, added.

aarontp (approved):
LGTM

PR #547, opened 2022-03-04
[Metawolf] Various bug fixes

giovannt0:
- Fix type checking for booleans 
- `set -a` now tells you if a parameter is optional or not
- Reorder the status of execution / timestamp of in-use recipes for `show -s`, which was incorrectly ordered and was missing entries for not-yet-run recipes
- Change `clear` to actually clear the screen


dftimewolf/metawolf/metawolf.py
```diff
@@ -257,8 +257,12 @@ def do_set(self, args: argparse.Namespace, _: Any) -> None:
           # Session and recipe settables are not in scope.
           continue
         try:
-          value = input('Set value for parameter {0:s} ({1:s}: {2:s}. Press '
-                        '"enter" to skip or ctrl+d to abort.): '.format(
+          value = input('Set value for{0:s}parameter {1:s} ({2:s}: {3:s}. '
```
tomchop, line 260: Are there spaces missing around the `{0:s}`?

dftimewolf/metawolf/metawolf.py
```diff
@@ -962,7 +967,7 @@ def UpdateSessionSettable(
 
     prev_value = settable.GetValue()
     value = utils.CastToType(updated_value, settable.type)
-    if not value:
+    if value is None:
```
tomchop, line 970: What if we're trying to clear a parameter and set it to `None`?

tests/metawolf/utils.py
```diff
@@ -86,7 +86,7 @@ def testStr2Bool(self) -> None:
   def testGetType(self) -> None:
     """Test that the type of a string is inferred correctly."""
     self.assertEqual(bool, utils.GetType('y'))
-    self.assertEqual(int, utils.GetType('0'))
+    self.assertEqual(bool, utils.GetType('0'))
```
tomchop, line 89: What if a recipe takes an int that can be 0?

tomchop (changes requested):

dftimewolf/metawolf/metawolf.py
giovannt0: No, this is on purpose

tests/metawolf/utils.py
giovannt0:
```
$b552a4$aws_forensics> set -p boot_volume_size 0
boot_volume_size: 50 --> 0
```

Not a problem

dftimewolf/metawolf/metawolf.py
giovannt0: Also fine
```
$b552a4$aws_forensics> reload
remote_profile_name: dasd --> None
remote_zone: asdasd --> None
incident_id: sdads --> None
all_volumes: True --> False
boot_volume_size: 0 --> 50
```

tomchop (approved):

PR #551, opened 2022-03-07
Remove log propagation from child to root

ramo-j:
This removes duplicate log entries when some modules are used:

Before:
```
[2022-03-07 03:37:45,074] [dftimewolf          ] INFO     Running module: LocalFilesystemCopy
[2022-03-07 03:37:45,074] [LocalFilesystemCopy ] INFO     <path> -> /tmp/dftimewolf_local_fs0t8tq4ta
2022-03-07 03:37:45 [INFO] <path> -> /tmp/dftimewolf_local_fs0t8tq4ta
[2022-03-07 03:37:45,079] [LocalFilesystemCopy ] SUCCESS  <path> was compressed into /tmp/dftimewolf_local_fs0t8tq4ta/<output>.tgz
2022-03-07 03:37:45 [SUCCESS] <path> was compressed into /tmp/dftimewolf_local_fs0t8tq4ta/<output>.tgz
[2022-03-07 03:37:45,079] [dftimewolf          ] INFO     Module LocalFilesystemCopy finished execution

```

After:
```
[2022-03-07 03:42:41,743] [dftimewolf          ] INFO     Running module: LocalFilesystemCopy
[2022-03-07 03:42:41,743] [LocalFilesystemCopy ] INFO     <path> -> /tmp/dftimewolf_local_fs76lmm6m2
[2022-03-07 03:42:41,748] [LocalFilesystemCopy ] SUCCESS  <path> was compressed into /tmp/dftimewolf_local_fs76lmm6m2/<output>.tgz
[2022-03-07 03:42:41,748] [dftimewolf          ] INFO     Module LocalFilesystemCopy finished execution
```

PR #554, opened 2022-03-07
Revert typo fix

ramo-j:
As per [0], reverting a typo fix.

[0] https://github.com/log2timeline/dftimewolf/pull/551#pullrequestreview-901478252


itsmvd (commented):
IMHO this is still an unclear print statement. Saying something like "Unable to get information for flow {flow_id}" would be better for user experience.

itsmvd (approved):

PR #555, opened 2022-03-14
Added in AWS->GCP E2E test

### LINKED ISSUE #457: e2e tests for AWS recipes
It would be good if we had e2e tests for the following modules:

* aws_disk_to_gcp
* aws_turbinia_ts

ramo-j:
Added in E2E test that covers copying disks from AWS to GCP via the following modules:
* AWSVolumeSnapshotCollector
* AWSSnapshotS3CopyCollector
* S3ToGCSCopy
* GCSToGCEImage
* GCEDiskFromImage

Closes #457

### DIFF
--- a/tests/e2e/aws_disk_forensics.py
+++ b/tests/e2e/aws_disk_forensics.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2020 Google Inc.
+# Copyright 2022 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""End to end test for the AWS -> GCP cloud forensics modules."""
+"""End to end test for the AWS -> GCP disk copy process."""
 import json
 import logging
 import os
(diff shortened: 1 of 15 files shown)


tests/e2e/aws_disk_forensics.py
```diff
@@ -0,0 +1,315 @@
...
+      'gcp_project': self.gcp_project_id
+    }
+    self.test_state.SetupModules()
+    self.test_state.RunModules()
+
+    # AWS Volume in count should equal GCE Disk out count, and be at least 1
+    self.assertGreaterEqual(
+        len(self.test_state.GetContainers(containers.AWSVolume)), 1)
+    self.assertEqual(len(self.test_state.GetContainers(containers.AWSVolume)),
+        len(self.test_state.GetContainers(containers.GCEDiskEvidence)))
```
tomchop, line 202: We're testing the internal state here, in e2e tests I think it would probably be better to leverage LCF to get the actual state of the remote project (and compare names that are there with the ones that dftimewolf claims to have created)

tomchop (changes requested):

tests/e2e/aws_disk_forensics.py
ramo-j: Unless I'm misunderstanding what you're asking, that's what's happening here. 199->202 is making sure that we have the same number of input disks in AWS as outputs in GCE according to the internal state (and also that it's not zero), then 204->207 is making sure those GCE disks actually exist via LCF.

tomchop: Yup I get this - what I'm suggesting is that we rely less on the internal state (given that this is e2e) and interact with the remote cloud API to confirm that the disks have been created and have the expected characteristics.

We do this with the [gcp_disk_forensics tests](https://github.com/log2timeline/dftimewolf/blob/main/tests/e2e/gcp_disk_forensics.py#L183)

ramo-j: Updated, as per OOB conversation.

tomchop (approved):

PR #562, opened 2022-03-22
Add empty result representation to _DownloadResults()

sydp:
Osquery flows that return no results are not reported (in PostProcessing for example) since they are based on results being stored in DataFrame containers.  This update injects an empty dataframe to capture the cases where osquery has no results.


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -737,7 +737,9 @@ def _DownloadResults(self,
 
     if not list_results:
       self.logger.info(f'No results for flow ID {str(flow)}')
-      return []
+      # Insert an empty dataframe to represent the flow completed without
+      # result.
+      return [pd.DataFrame()]
```
tomchop, line 742: Wouldn't returning an empty list make more sense here? Otherwise this means "there was a result, but no rows were returned"

tomchop (changes requested):

dftimewolf/lib/collectors/grr_hosts.py
sydp: Returning empty list here and added logic to the caller (Process()) to hande when the result had no rows.

tomchop (approved):

PR #568, opened 2022-03-29
WIP - Add support for query packs

sydp:
This PR adds support for query packs to the GRROsqueryCollector.

### DIFF
--- a/dftimewolf/lib/collectors/osquery.py
+++ b/dftimewolf/lib/collectors/osquery.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Collects osquery from the command line and the local file system."""
 
+import json
 import os
 from typing import Optional, List
 
@@ -9,12 +10,13 @@ from dftimewolf.lib.containers import containers
 from dftimewolf.lib.modules import manager as modules_manager
 from dftimewolf.lib.state import DFTimewolfState
 
+_ALL_PLATFORMS = ['darwin', 'freebsd', 'linux', 'windows']
 
 class OsqueryCollector(module.BaseModule):
   """Osquey query collector.
 
   Attributes:
-      osqueries (List[str]): list of osquery queries.
+      osqueries (List[containers.OsqueryQuery]): list of osquery containers.
   """
 
   def __init__(self,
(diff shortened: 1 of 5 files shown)


dftimewolf/lib/collectors/osquery.py
```diff
@@ -45,11 +47,95 @@ def _ValidateOsquery(self, query: str) -> bool:
...
     return query.upper().startswith('SELECT ')
 
+  def _ParsePlatforms(self, platforms: str) -> List[str]:
+    """Parse and normalise the platforms value from an osquery pack.
+
+    Arguments:
+      platforms: the platforms value from an osquery pack.
+
+    Returns:
+      a list of operating system platforms.
```
tomchop, line 57: Please specify what the possible values of the list are

dftimewolf/lib/collectors/osquery.py
```diff
@@ -45,11 +47,95 @@ def _ValidateOsquery(self, query: str) -> bool:
...
+  def _LoadOsqueryPack(self, path: str) -> None:
+    """Loads osquery from an osquery pack file.
+
+    Args:
+      path: the path to the JSON file.
+    """
+    with open(path, mode='r') as fd:
+      global_platform = []
+
+      query_pack = json.load(fd)
```
tomchop, line 85: Anything eblow this line can be unindented, doesn't need to have the filedescriptor open.

dftimewolf/lib/collectors/osquery.py
```diff
@@ -45,11 +47,95 @@ def _ValidateOsquery(self, query: str) -> bool:
...
+
+      # A 'global' platform value can be set at the root level
+      if 'platform' in query_pack:
+        global_platform = self._ParsePlatforms(query_pack.get('platform'))
+
+      for num, (name, entry) in enumerate(
+          query_pack.get('queries', {}).items()):
+        if not self._ValidateOsquery(entry.get('query', '')):
```
tomchop, line 93: Should we just access `entry['query']` and catch a KeyError? I feel it's safe to assume that the dicts will be well-constructed.

tomchop: Also, given that you're using it below, i would just store this in an intermediate variable.

dftimewolf/lib/collectors/osquery.py
```diff
@@ -45,11 +47,95 @@ def _ValidateOsquery(self, query: str) -> bool:
...
+        else:
+          platform = global_platform
+        self.osqueries.append(
+            containers.OsqueryQuery(
+                query=entry.get('query', ''),
+                name=name,
+                description=entry.get('description', ''),
+                platforms=platform))
+
+  def _LoadTextFile(self, path: str) -> None:
```
tomchop, line 109: I'd change the name of this to something that reflects that containers are being added, or at least specify in the docstring that "loading" here means actually creating attribute containers.

tests/lib/collectors/osquery.py
```diff
@@ -87,6 +87,55 @@ def testProcess(self, mock_exists) -> None:
...
+    """Tests the collector's Process() function with a Osquery Pack file."""
+    test_state = state.DFTimewolfState(config.Config)
+    osquery_collector = osquery.OsqueryCollector(test_state)
+    mock_exists.return_value = True
+
+    test_ok_data = """{
+      "platform": "darwin",
+      "queries": {
+        "query_1": {
+          "query" : "select * from launchd where path like '%System.plist';",
```
tomchop, line 101: Thisis supposed to be a string, but the JSON withing should still be represented correctly. Another way to do this would be to just write a dict and call `json.dumps` on it to make it a string. 

```suggestion
          "query": "select * from launchd where path like '%System.plist';",
```

tomchop (changes requested):

dftimewolf/lib/collectors/osquery.py
sydp: Done.

dftimewolf/lib/collectors/osquery.py
sydp: Done.

dftimewolf/lib/collectors/osquery.py
sydp: Done.

dftimewolf/lib/collectors/osquery.py
sydp: Done.

tests/lib/collectors/osquery.py
```diff
@@ -87,6 +87,55 @@ def testProcess(self, mock_exists) -> None:
...
+        "query_2": {
+          "query" : "select * from test where path like '%user32.dll';",
+          "interval" : "3600",
+          "version": "1.4.5",
+          "platform": "windows",
+          "description" : "description",
+          "value" : "Artifact used by this malware"
+        }
+      }
+    }"""
```
tomchop, line 116: Please fix the spaces throughout this whole blob (that's why I suggested using dumps in the first place heh)

tomchop (changes requested):

tests/lib/collectors/osquery.py
sydp: My bad, thought I updated but guess the update didn't get pushed up.  Should be fixed now (pending PR actions).

tomchop (approved):

PR #569, opened 2022-03-29
Internal module logging

tomchop:
This pull request enables modules to log stats that will later might be exported / logged by the main dfTimewolf process.


dftimewolf/lib/state.py
```diff
@@ -217,6 +236,23 @@ def StoreContainer(self, container: "interface.AttributeContainer") -> None:
     with self._state_lock:
       self.store.setdefault(container.CONTAINER_TYPE, []).append(container)
 
+  def StoreStats(self, statsentry: StatsEntry) -> None:
+    """Thread-safe method to store stats in the state's stats store.
+
+    Args:
+      statsentry: The stats object to store.
+    """
+    with self._state_lock:
```
ramo-j, line 245: This is the same mutex that Get/Store Containers, there's no possibility of a deadlock here is there?

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -292,6 +292,20 @@ def CleanUpPreflights(self) -> None:
...
+    if not statentries:
+      logger.info('No statistics collected during execution.')
+
+    logger.info(f'{len(statentries)} stat entries collected during execution.')
+    for entry in statentries:
+      logger.debug(f'[{entry.module_name} ({entry.module_type})] {entry.stats}')
+
+  def ExportStats(self) -> None:
+    """Exports collected stats if existing. Default behavior is to log."""
+    self.PrintStats()
```
ramo-j, line 307: Is there a plan to extend this to publish stats to a file, or external service for collection?

dftimewolf/lib/state.py
```diff
@@ -217,6 +236,23 @@ def StoreContainer(self, container: "interface.AttributeContainer") -> None:
...
+      self.stats_store.append(statsentry)
+
+  def GetStats(self) -> List[StatsEntry]:
+    """Get stats entries that have been stored in the state.
+
+    Returns:
+       The stats objects stored in the state's stats store.
+    """
+    return self.stats_store
```
ramo-j, line 254:
```suggestion
    with self._state_lock:
      return self.stats_store
```

tests/lib/state.py
```diff
@@ -410,6 +410,21 @@ def testContainerDedupe(self):
...
+    test_state.SetupModules()
+    test_state.RunModules()
+    stats = test_state.GetStats()
+    self.assertEqual(len(stats), 2)
+    self.assertIsInstance(stats[0], state.StatsEntry)
+    self.assertEqual(stats[0].module_name, 'DummyModule1')
+    self.assertEqual(stats[1].module_name, 'DummyModule2')
+    self.assertEqual(stats[0].stats, {'random_key1': 'random_value1'})
+    self.assertEqual(stats[1].stats, {'random_key2': 'random_value2'})
+
```
ramo-j, line 427: I'm a big fan of testing for failures as well, should we add some in? I can see that setting a stat with a key that isn't a string should fail, for example.

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -292,6 +292,20 @@ def CleanUpPreflights(self) -> None:
     """Calls the preflight's CleanUp functions."""
     self._state.CleanUpPreflights()
 
+  def PrintStats(self) -> None:
+    """Prints collected stats if existing."""
+    statentries = self._state.GetStats()
```
ramo-j, line 297: Nit: `state_entries` instead? Because I first read this as `state_ntries`

dftimewolf/lib/state.py
tomchop: Added your acquire lock suggestion

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Yes absolutely, and they should override this very `ExportStats` method

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Great suggestion, done.

ramo-j (approved):
LGTM

PR #574, opened 2022-04-01
Bugfix - Check evidence_disk is not null in ForensicsVM container

ramo-j:
Currently, if the ForensicsVM has no evidence_disk attribute, turbinia gcp attempts to create a GCEEvidenceDisk container regardless, throwing an exception. This change corrects that.


sydp (approved):
LGTM

Issue #575, opened 2022-04-01
Github PR actiosn failing with `ModuleNotFoundError: No module named 'setuptools_rust'`


ramo-j:
Example: https://github.com/log2timeline/dftimewolf/runs/5748380736?check_suite_focus=true

Digitalisx:
I forgot to link the issue with PR. 
So, it couldn't be closed automatically when merged.
If there is no other problem, I think the issue can be closed.
Thank you for raising the issue! 😊

PR #583, opened 2022-04-10
Fix: install dependencies error with pip and setuptools update.

Digitalisx:
Hello, everyone in the community! 🙂

We have identified some errors in building and testing dftimewolf.
To resolve this error, we had to upgrade the version of pip and setuptools.
If there is no big problem, the check will be carried out normally.

If you have any additional comments or feedback, please leave a thread!

**Related Issue**
- [x] #575


sydp (approved):
Thanks for the contribution @Digitalisx.  LGTM.

PR #585, opened 2022-04-12
Add pytype github action

ramo-j:
Added in a github action for pytype type checking. Also, code updates for the checking to pass.


dftimewolf/lib/enhancers/timesketch.py
```diff
@@ -41,16 +41,17 @@ class TimesketchEnhancer(module.BaseModule):
...
                state: "dftw_state.DFTimewolfState",
                name: Optional[str]=None,
                critical: bool=False):
     super(TimesketchEnhancer, self).__init__(
         state, name=name, critical=critical)
-    self.timesketch_api: "client.TimesketchApi"
-
     self._aggregations_to_skip = []  # type: List[str]
-    self._formatter: utils.FormatterInterface
     self._include_stories = False
```
itsmvd, line 55: Lets be explicit and consistent and also specify these types, just like we're doing for types that are a List.
So include # type: bool etc.

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -45,11 +47,11 @@ def __init__(self,
     self._analyzers = []  # type: List[str]
     self.wait_for_timelines = False
     self.host_url = None
-    self.sketch = None  #type: ts_sketch
+    self.sketch = None
```
itsmvd, line 50: Why are we removing this type? Would be good to add it, should it be/have been ts_sketch.Sketch?

tests/lib/collectors/gce_disk_copy.py
```diff
@@ -388,12 +388,12 @@ def testProcess(self,
     collector.PreProcess()
     conts = test_state.GetContainers(collector.GetThreadOnContainerType())
     for d in conts:
-      collector.Process(d)
+      collector.Process(d)  # pytype: disable=wrong-arg-types
```
itsmvd, line 391: Can we add a note why we have to disable pytype here? Just because this has to be disabled throughout this file, be good to have a note about that for the first time you're disabling it for conts/d.

itsmvd (changes requested):

dftimewolf/lib/enhancers/timesketch.py
ramo-j: Done.

dftimewolf/lib/exporters/timesketch.py
ramo-j: Added back.

tests/lib/collectors/gce_disk_copy.py
ramo-j: Done.

itsmvd (approved):

PR #598, opened 2022-05-20
Added in the CursesDisplayManager and tests

ramo-j:
Created a curses display management class to handle displaying DFTW state via curses, rather than simply lines of log output. 

This is the first of 2 planned PRs, split up in the interests of PR size. This PR just creates the CursesDisplayManager class. A later PR will actually hook it up for usage.

### DIFF
--- a/dftimewolf/cli/curses_display_manager.py
+++ b/dftimewolf/cli/curses_display_manager.py
@@ -82,15 +82,12 @@ class Module:
 
 class Message:
   """Helper class for managing messages."""
-  source: str = ''
-  content: str = ''
-  is_error: bool = False
 
   def __init__(self, source: str, content: str, is_error: bool = False) -> None:
     """Initialise a Message object."""
-    self.source = source
-    self.content = content
-    self.is_error = is_error
+    self.source: str = source
+    self.content: str = content
+    self.is_error: bool = is_error
 
   def Stringify(self, source_buff_len: int = 0, colourise: bool = False) -> str:
     """Returns an CursesDisplayManager friendly string of the Message."""
@@ -101,6 +98,7 @@ class Message:
     reset_code = '\u001b[0m' if self.is_error and colourise else ''
 
     return f'[ {self.source:{pad}} ] {colour_code}{self.content}{reset_code}'
+
 
 class CursesDisplayManager:
   """Handles the curses based console output, based on information passed in.
@@ -139,7 +137,7 @@ class CursesDisplayManager:
     self._recipe_name = recipe
 
   def SetException(self, e: Exception) -> None:
-    """Set a Exception to be included in the display."""
+    """Set an Exception to be included in the display."""
     self._exception = e
 
   def SetError(self, module: str, message: str) -> None:
--- a/dftimewolf/cli/curses_display_manager.py
+++ b/dftimewolf/cli/curses_display_manager.py
@@ -30,7 +30,13 @@ class Module:
                name: str,
                dependencies: List[str],
                runtime_name: Optional[str] = None):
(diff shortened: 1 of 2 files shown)


dftimewolf/cli/curses_display_manager.py
```diff
@@ -0,0 +1,290 @@
...
+      # Messages
+      curr_line += 1
+      self._stdscr.addstr(curr_line, 0, ' Messages:')
+      curr_line += 1
+
+      message_space = y - 4 - curr_line
+      start = len(self._messages) - message_space
+      start = 0 if start < 0 else start
+
+      for m in self._messages[start::]:
```
tomchop, line 255: Why the `::` here? If we need to create a slice copy, it would be nice to add a comment detailing why.

dftimewolf/cli/curses_display_manager.py
```diff
@@ -0,0 +1,290 @@
...
+    self.source: str = source
+    self.content: str = content
+    self.is_error: bool = is_error
+
+  def Stringify(self, source_buff_len: int = 0, colourise: bool = False) -> str:
+    """Returns an CursesDisplayManager friendly string of the Message."""
+    pad = (len(self.source) if len(self.source) > source_buff_len
+        else source_buff_len)
+
+    colour_code = '\u001b[31m' if self.is_error and colourise else ''
```
tomchop, line 97: So far we've used americanized English, e.g. `color`. Would you mind adjusting this? Not a hill I'm willing to die on but I guess it's better to make it consistent throughout the codebase.

dftimewolf/cli/curses_display_manager.py
```diff
@@ -0,0 +1,290 @@
...
+
+    return f'[ {self.source:{pad}} ] {colour_code}{self.content}{reset_code}'
+
+
+class CursesDisplayManager:
+  """Handles the curses based console output, based on information passed in.
+  """
+
+  def __init__(self) -> None:
+    """Intiialises the CursesDisplayManager."""
```
tomchop, line 108:
```suggestion
    """Intializes the CursesDisplayManager."""
```

dftimewolf/cli/curses_display_manager.py
```diff
@@ -0,0 +1,290 @@
...
+            f'       {n}: {t["status"].value} ({t["container"]})')
+
+    return [module_line] + thread_lines
+
+  def SetStatus(self, status: Status) -> None:
+    """Set the status of the module."""
+    if self.status not in [Status.ERROR, Status.COMPLETED, Status.CANCELLED]:
+      self.status = status
+
+  def SetThreadState(self, thread: str, status: Status, container: str) -> None:
```
tomchop, line 66: Could you have a bit more detailed (e.g. adding an `Args:` sections) docstrings here and throughout the PR? I'm struggling to understand what e.g. the container str is used for, or should be.

tomchop (changes requested):

dftimewolf/cli/curses_display_manager.py
ramo-j: Added.

dftimewolf/cli/curses_display_manager.py
ramo-j: A relic of an earlier design, where I had the messages printed in reverse order. Fixed.

dftimewolf/cli/curses_display_manager.py
ramo-j: As much as it pains me, done. I think I got it all, but since it's my native, I might have missed some.

tomchop (approved):

PR #606, opened 2022-05-30
Second stage of curses UI implementation

ramo-j:
Following on from #598 this PR is the second stage of the curses user interface. This takes the CursesDisplayManager from that PR and actually hooks it up for usage.

The previous UI (log message stream) can still be used by setting an env var `DFTIMEWOLF_NO_CURSES`


tomchop (changes requested):
Instead of having all these if branches, wouldn't it be cleaner if we had a completely separate class dealing with the curses stuff? WDYT?

dftimewolf/lib/module.py
```diff
@@ -117,6 +118,16 @@ def ModuleError(self, message: str, critical: bool=False) -> None:
...
 
+  def PublishMessage(self, message: str, is_error: bool = False) -> None:
+    """Sends a message to the CDM to disaply.
+
+    Args:
+      message: The message content.
+      is_error: True if the message is an error message, False otherwise."""
+    if self.state.cdm:
+      self.state.cdm.EnqueueMessage(self.name, message, is_error)
+    self.logger.info(message)
```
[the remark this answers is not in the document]
ramo-j: Yeah that did cross my mind, but I opted away from it. There's no reason we couldn't have an optional logging level parameter, but I'm not sure it's necessary.

ramo-j:
> Instead of having all these if branches, wouldn't it be cleaner if we had a completely separate class dealing with the curses stuff? WDYT?

I'm definitely interested in cleaning up the sprawling `if self.cdm:`'s somehow. I'm not sure what a separate class look like though, what did you have in mind?

dftimewolf/lib/module.py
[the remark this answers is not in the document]
tomchop: I don't really know how that will look like in the end result so it's hard for me to tell. But the `success` statements should be kept, since they are the ones that indicate to the user where to find things (e.g. GRR flow IDs, timesketch URLs, etc.). If you're telling me that we only keep 5% of the lines we had before, then I think I can survive without any fancy coloring :)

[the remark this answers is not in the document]
ramo-j: No logging is getting removed (it all still ends up in the debug log), but I think we could use any "success" log as an indicator that it should use PublishMessage instead.

ramo-j:
> What I imagined was a class (e.g. extending `Module` and `State`) that overrides functions where the if branches exist. This might require significant refactoring (e.g. putting everything in functions, or improve the separation of concerns and really separate the interface from the module and state classes, but this would require a major redesign), and depending on how much we're relying on the superclass, it might deduplicate work every time we want to change something in those classes.
> 
> You're best placed to know how much this curses code will get in the way of other functional changes, so I'll leave it up to you. I think it's also OK to merge this now, but really consider whether we want to do this cleanup or not, and document reasons why.

`State` yeah I can get behind, but I don't think it makes sense for `Module`. The change in Module is pretty small, and it also creates a `PublishMessage` method for modules to use, which should be available to any module, no matter what the display looks like. 

I'll give it a crack for State, see what it looks like and get back to you.

> all this of course if we want to keep both interfaces for some reason.

Metawolf comes to mind, I don't know enough about how that operates to know if it would interfere. But also in case something breaks in a way I haven't thought of, I think it makes sense to be able to fallback to the old mode via the Env var, at least until we're super confident in the change.

tomchop (dismissed):

tomchop:
OK, happy to approve this now. up to you whether you want to give State a crack now or in a future PR.

ramo-j:
Subclassing `state` makes this look much cleaner, including a side effect of making `_RunModuleThread` much easier to understand as well, so I think it's a better method than the previous sprawling `if's`. 

Unfortunately, it has blown out the size of this PR by an extra ~400-500 lines. A lot of that is in a new test for the subclass.

dftimewolf/lib/module.py
```diff
@@ -64,7 +64,7 @@ def SetupLogging(self, threaded: bool = False) -> None:
         threaded=threaded))
     self.logger.addHandler(file_handler)
 
-    if not self.state.cdm:
+    if not isinstance(self.state, state_lib.DFTimewolfStateWithCDM):
```
tomchop, line 67: Can we have a boolean / property in `state` that indicates whether or not it's a CDM state? So we can have 

```python
if not self.state.iscursed:
```

or something.

Same goes for other places where `isinstance` is called

dftimewolf/lib/state.py
```diff
@@ -614,3 +600,157 @@ def CheckErrors(self, is_global: bool=False) -> None:
...
+    self.cursesdm.UpdateModuleStatus(module.name, cdm.Status.PENDING)
+
+  def _RunModuleProcess(self, module: BaseModule) -> None:
+    """Runs Process of a single module.
+
+    Args:
+      module: The module to run Process() on."""
+    self.cursesdm.UpdateModuleStatus(module.name, cdm.Status.PROCESSING)
+    module.Process()
+    self.cursesdm.UpdateModuleStatus(module.name, cdm.Status.COMPLETED)
```
tomchop, line 662: This looks so much better! 👍

dftimewolf/lib/state.py
```diff
@@ -614,3 +600,157 @@ def CheckErrors(self, is_global: bool=False) -> None:
...
+    CursesDisplayManager status update methods.
+
+    Args:
+      process: A callable method: Process, belonging to a ThreadAwareModule.
+      container: The Container being processed by the thread.
+      module_name: The runtime name of the module."""
+
+    thread_id = threading.current_thread().getName()
+    container_name = (container.name  # type: ignore[attr-defined]
+        if hasattr(container, 'name') else str(container))
```
tomchop, line 738: Why not just encode this logic `container.__str__`? I like the idea that each container defines how it's being displayed to a user.

tomchop (approved):
The changes look great, thanks!

dftimewolf/lib/state.py
ramo-j: Makes sense.

dftimewolf/lib/module.py
ramo-j: Going the subclass method, I didn't want to have anything referring to curses in the base class. But, we can for sure have a bool `stdout_log` which accomplished the same thing. 

As for the other references, I've refactored `PublishMessage` to acheive the same goal of keeping curses pieces away from state base class and `module`.

ramo-j:
Latest changes dismissed the approval. Adding local timezone approver, since I think everything has been covered now.

jonathan-greig (approved):
LGTM based on previous approval.

PR #607, opened 2022-05-31
Added delay between GCE launch and evidence disk attachment

ramo-j:
If an evidence disk is attached too close to instance launch, and the evidence disk is a boot disk from the copied VM, the VM can fail to pick the correct disk to launch from, since both disks are equally correct, as far as the VM is concerned. Adding a delay to attaching evidence disks will ensure that the correct disk is booted from.

### DIFF
--- a/dftimewolf/lib/processors/gce_forensics_vm.py
+++ b/dftimewolf/lib/processors/gce_forensics_vm.py
@@ -154,7 +154,8 @@ class GCEForensicsVM(module.BaseModule):
 
     disks = self.state.GetContainers(containers.GCEDisk)
 
-    # Sleep until status is RUNNING before attaching disks. Possible status values:
+    # Sleep until status is RUNNING before attaching disks. Possible status
+    # values:
     # https://cloud.google.com/compute/docs/reference/rest/v1/instances/get
     if disks and self.analysis_vm.GetPowerState() != 'RUNNING':
       self.logger.info('Pausing 10 seconds to allow OS to boot before attaching'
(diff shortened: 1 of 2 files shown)


dftimewolf/lib/processors/gce_forensics_vm.py
```diff
@@ -153,6 +154,11 @@ def Process(self) -> None:
 
     disks = self.state.GetContainers(containers.GCEDisk)
 
+    if disks:
+      self.logger.info('Pausing 30 seconds to allow OS to boot before attaching'
+          ' evidence disks')
+      time.sleep(30)
```
sa3eed3ed, line 160:
```suggestion
    # Sleep until status is RUNNING before attaching disks. Possible status values:
    # https://cloud.google.com/compute/docs/reference/rest/v1/instances/get
    if disks and self.analysis_vm.GetPowerState() != 'RUNNING':
      self.logger.info('Pausing 10 seconds to allow OS to boot before attaching'
          ' evidence disks')
      time.sleep(10)
```

sa3eed3ed (changes requested):
thank you, just added a check on the instance status and to make the wait shorter

sa3eed3ed (dismissed):

dftimewolf/lib/processors/gce_forensics_vm.py
```diff
@@ -153,6 +154,13 @@ def Process(self) -> None:
 
     disks = self.state.GetContainers(containers.GCEDisk)
 
+    # Sleep until status is RUNNING before attaching disks. Possible status values:
```
sa3eed3ed, line 157:
```suggestion
    # Sleep until status is RUNNING before attaching disks. Possible values:
```

sa3eed3ed: to solve the line too long pylint error

sa3eed3ed (approved):

sa3eed3ed (approved):

PR #608, opened 2022-05-31
Add optional prefix to image names

tomchop:


dftimewolf/lib/exporters/gce_image_from_disk.py
```diff
@@ -84,7 +89,9 @@ def Process(self, container: containers.GCEDisk) -> None:
     # use a class member due to an underlying thread safety issue in
     # httplib2: https://github.com/googleapis/google-cloud-python/issues/3501
     project = gcp_project.GoogleCloudProject(self.dest_project)
-    image = project.compute.CreateImageFromDisk(source_disk)
+    timestamp = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+    image_name = f'{self.name_prefix}-{source_disk.name}-{timestamp}'
```
ramo-j, line 93: The name limit here could easily be blown out, as the max is 63 characters before it gets trimmed by LCF. And it's not unusual, when interacting with earlier modules in a recipe that `evidence`, `copy` and a 5(?) char random string with `-` between them all are added to the name before it makes it to this module. Not that I'm sure what we'd do to trim this, maybe a 2 character year `%y` and drop the seconds `%S` as a start.

ramo-j (dismissed):
A note about name length I don't have a great solution for, and noting the pending linter appeasement, otherwise LGTM.

dftimewolf/lib/exporters/gce_image_from_disk.py
tomchop: I'll just drop the timestamp altogether, as the time of the image creation can still be found in the GCP backend.

ramo-j (approved):

PR #610, opened 2022-06-02
Don't fail the GCEForensicsVM module if a evidence disk can't be attached.

### LINKED ISSUE #609: Jenkins job dfTimewolf-libcloudforensics_e2e is currently failing

ramo-j:
Currently, if the GCEForensicsVM fails to attach a disk because it is already attached to another instance, there is an uncaught `RuntimeError` thrown. This can happen if evidence disks are copied from the same project where the forensics VM is created (as seen in the E2E test.) This change checks if that is the case, and prints a warning, but continues.

Fixes #609

### DIFF
--- a/dftimewolf/lib/processors/gce_forensics_vm.py
+++ b/dftimewolf/lib/processors/gce_forensics_vm.py
@@ -164,11 +164,19 @@ class GCEForensicsVM(module.BaseModule):
     for d in disks:
       if d.project != self.project.project_id:
         continue
-      self.logger.info(f'Attaching {d.name} to {analysis_vm_name}')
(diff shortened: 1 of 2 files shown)


ramo-j:
Unsure what the story is with `Build documentation pages / build (pull_request)`
```
Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_fc85c860-81f5-42d3-80f0-aabb362171d4/eca28826-0cc7-45d9-b029-91456ad4d198.tar.gz. return code: 2.
```

jonathan-greig (approved):

PR #612, opened 2022-06-03
Change AttributeContainer "attributes" to a Dict and rename

### LINKED ISSUE #611: Change AttributeContainer "attributes" to a Dict and rename
The container base class `dftimewolf.lib.containers.interface.AttributeContainer` has a generic `attributes` attribute for subclasses to use as "A list of generic attributes that can be used for passing metadata between collection/processing module and output modules."

This attribute is currently a list but looking at current uses of this feature, it would be more useful as a Dict. 

Also the term "attributes" is overloaded in this context and leads to confusing code.

As this feature is only used in a couple of modules I suggest renaming it to `container_meta` or similar to make it clearer what this feature should be used for.

jonathan-greig:
Fixes #611

### DIFF
--- a/dftimewolf/lib/containers/interface.py
+++ b/dftimewolf/lib/containers/interface.py
@@ -17,23 +17,23 @@ class AttributeContainer():
   and private class members are not to be serialized.
 
   Attributes:
-    attributes: A list of generic attributes that can be used for passing
+    container_meta: A dict of container metadata that can be used for passing
       metadata between collection/processing module and output modules.
   """
   CONTAINER_TYPE = None  # type: str
-  attributes = []  # type: List[Dict[str, Any]]
+  container_meta = {}  # type: Dict[str, Any]
 
-  def __init__(self, attributes: Optional[List[Dict[str, Any]]] = None):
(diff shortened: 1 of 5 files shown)


dftimewolf/lib/containers/interface.py
```diff
@@ -17,23 +17,23 @@ class AttributeContainer():
   and private class members are not to be serialized.
 
   Attributes:
-    attributes: A list of generic attributes that can be used for passing
+    container_meta: A dict of container metadata that can be used for passing
```
tomchop, line 20: Can we just call this `meta` or `metadata`? `container_meta` sounds a bit redundant. (especially given that in the code we might call it `container.container_meta`)

tomchop (commented):
One nit, otherwise LGTM. Feel free to merge once that sole comment is adressed.

dftimewolf/lib/containers/interface.py
jonathan-greig: Done.

ramo-j (approved):

PR #624, opened 2022-06-25
Fix GRRFileCollector to use TSK API input for GRR

coperni:
By [default](https://github.com/google/grr/blob/master/grr/proto/grr_response_proto/flows.proto#L1852) the PathType for the GRR File Finder flow uses `OS`. When TSK is enabled via the command line this option is passed into the GRR client's flow arguments.


tomchop (approved):
Thanks for your contribution!

tomchop (approved):

PR #625, opened 2022-06-27
refactor TurbiniaProcess method

sa3eed3ed:
refactor the TurbiniaProcess function, move the request polling logic to a new function and make it optional (default to True not break current clients).


sa3eed3ed:
> I'm not sure I understand the purpose of this CL, the code seems to behave exactly like before? Is the intent here to be able to override TurbiniaWait with our own class?

This is to be able to use the TurbiniaWait and TurbiniaStart logic independently and refactor the child module TurbiniaGCPProcessor into two modules, one starts a request and one wait for it to finish, accordingly have the ability to add a recipe  to resume an existing Turbinia job

tomchop (approved):

Issue #486, opened 2021-12-08
Contributor's guide


tomchop:
We want a contributor's guide / help text that is a bit more verbose than what we have now, namely on code idiosyncracies:

* How to write a good recipe + style
* How to write a good module (decoupling) + style
* When to add a new ArtifactContainer class
* How big do we want PRs to be / how to articulate contributions.

Digitalisx:
I'm looking forward to this issue!

If the issue is resolved one day, the link to the contribution guide or code style that exists in the current document is invalid and will need to be updated.

https://github.com/log2timeline/dftimewolf/blob/02aa4cc778fa58baa7adc197947e321807e319e7/docs/developers-guide.md?plain=1#L41

https://github.com/log2timeline/dftimewolf/blob/02aa4cc778fa58baa7adc197947e321807e319e7/docs/developers-guide.md?plain=1#L46

tomchop:
We should revive this now that we have more people contributing.

PR #626, opened 2022-07-05
Added in module writing guide

ramo-j:
As per #486, a doc covering how to write a module at a very basic level.

### DIFF
--- a/docs/module-writing-basics.md
+++ b/docs/module-writing-basics.md
@@ -1,0 +1,195 @@
+# How to write a DFTimewolf Module
+
+The purpose of this guide is to walk a contributor through the basics of how to write a DFTimewolf module. Our example will perform the following:
+
+* Receive filepaths of files on the local filesystem
+* Hash the files and check against a list of hashes
+* Move matching files to an output directory
+* Keep only matching files in the state for processing by a later module
+
+## 0 - An empty template
+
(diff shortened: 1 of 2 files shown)


docs/module-writing-basics.md
```diff
@@ -0,0 +1,195 @@
...
+\_\_init__ is simple enough, we only need to declare the class members:
+
+```python
+	def __init__(self, state, name, critical):
+		super(FileRegexModule, self).__init__(state, name=name, critical=critical)
+		self.hashes = []
+		self.destination_dir = ''
+```
+
+`SetUp()` is called by the orchestration with parameters as defined by the recipe file. Our module is going to receive two parameters: a comma separated list of files and a comma separated list of hashes.
```
tomchop, line 128: Might be worth specifying that they're getting comma-separated strings instead of lists because they're supposed to be passed in by humans through the CLI (and not directly from a recipe)

docs/module-writing-basics.md
```diff
@@ -0,0 +1,195 @@
...
+			if p:
+				filename = os.path.basename(p)
+				self.state.StoreContainer(containers.File(name=filename, path=p))
+```
+
+## 3 - Processing
+
+Now we arrive at the heavy lifting of our module, we have 3 methods remaining to implement.
+
+PreProcess is used for any actions that aren't considered part of the `SetUp()` but also need to be taken before processing. In an example that operates on a cloud platform, you could use this to create IAM permissions needed to perform the work.
```
tomchop, line 146:
```suggestion
`PreProcess` is used for any actions that aren't considered part of the `SetUp()` but also need to be taken before processing. In an example that operates on a cloud platform, you could use this to create IAM permissions needed to perform the work. 
```

tomchop (dismissed):
A few nits; thanks a lot for writing this!

docs/module-writing-basics.md
ramo-j: Done.

PR #627, opened 2022-07-06
Redirect stdout to stderr for argparse

ramo-j:
Currently, when passing `-h` to a recipe, argparse sends the output to stdout. In the context of the curses UI we'd prefer it be sent to stderr, so it gets captured and displayed to the user.


ramo-j:
Actions failing for dependencies, so I assume https://github.com/log2timeline/dftimewolf/pull/623 is a blocker here.

jonathan-greig (approved):

PR #629, opened 2022-07-07
Surface the turbinia request ID to the user via PublishMessage

ramo-j:
Surface the turbinia request ID to the user via PublishMessage.

Required a small amount of refactoring - moving `TurbiniaStart` and `TurbiniaWait` out of `TurbiniaProcess` into the child classes (and removing `TurbiniaProcess` as it's not used any more as a result.)

### DIFF
--- a/dftimewolf/lib/processors/turbinia_artifact.py
+++ b/dftimewolf/lib/processors/turbinia_artifact.py
@@ -86,7 +86,9 @@ class TurbiniaArtifactProcessor(TurbiniaProcessorBase,
     evidence_ = evidence.CompressedDirectory(
         compressed_directory=container.path, source_path=container.path)
     try:
-      task_data, _ = self.TurbiniaProcess(evidence_)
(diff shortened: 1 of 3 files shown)


sa3eed3ed:
This is LGTM and can be submitted with no conflicts. 
FYI I have a new version (internal fork) where:
- the turbinia_gcp module is split into a Start module and a Wait module
- it is possible to resume an existing Turbinia request with a new recipe that only uses the wait module

I was planing to clean-up the turbinia_base.py module when bringing in the internal fork but removing TurbiniaProcess function now won't harm.

I will merge this code here a bit later for the sake of better testing, **so please make sure we do not do redundant work**

sa3eed3ed (approved):

ramo-j:
> ramo-j I will merge this code here a bit later for the sake of better testing,
Roger that @sa3eed3ed, I'll let you hit the merge button on this PR when you're ready

PR #630, opened 2022-07-08
Fix for downloading grr timeline flow results

jkppr:
With this modification to the `_DownloadFiles()` function, the grr collector now checks if a given GRR flow is a TimelineFlow and then proceeds to download the resulting TimelineBody file instead of the empty FilesArchive.  This will allow dftimewolf recipes to also process timeline results from GRR that have been collected earlier e.g. via the GRR interface.

### DIFF
--- a/dftimewolf/lib/collectors/grr_hosts.py
+++ b/dftimewolf/lib/collectors/grr_hosts.py
@@ -354,7 +354,10 @@ class GRRFlow(GRRBaseModule, module.ThreadAwareModule):
       os.makedirs(client_output_folder)
 
     if is_timeline_flow:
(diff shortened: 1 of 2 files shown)


jkppr:
> LGTM! Just one thing - can we add a `self.PublishMessage()` (or at least, `self.logger.info()`) statement so the output file can be grabbed by the user if anything else ends up failing?

As discussed, this should already be covered by the Process function of the module: https://github.com/log2timeline/dftimewolf/blob/f481aec89a6cca384dd35a8735ee2bb4c3328c03/dftimewolf/lib/collectors/grr_hosts.py#L632

tomchop (approved):

PR #635, opened 2022-07-18
new lcf version

sa3eed3ed:


tomchop (changes requested):
Please generate the requirements file through Pipenv (so that a pipfile.lock is also synced)

jleaniz:
I think we can close this PR?

PR #636, opened 2022-07-18
Move to Poetry and pyproject.toml

tomchop:
Poetry (https://python-poetry.org/) is yet another python dependency / virtualenv manager, but it provides significant advantages over pipenv:

* It can actually update single dependencies. This is something we do quite often with turbinia, grr, libcloudforensics, timesketch, e.g. https://github.com/log2timeline/dftimewolf/pull/635
  * pipenv can't, and probably won't be able to, do this https://github.com/pypa/pipenv/issues/2412, https://github.com/pypa/pipenv/pull/4988
* It can also take care of building and pushing to pypi (which we don't currently do, but that would be an option to explore). 

It can still generate requirements.txt files for people who want to fall back to just using pip.

Edit: I also upgraded `pylint`, which was quite outdated, and copied / pasted the log2timeline/plaso pylintrc file. We need to go over the error messages / codebase to see what warnings we want to keep and what new things to fix: https://github.com/log2timeline/dftimewolf/issues/637


sa3eed3ed (dismissed):

tomchop:
I'll hold off on @ramo-j's explicit LGTM on this before merging

ramo-j:
I notice the update to pylint is causing quite a number of new linting errors. My preference would be to fix them in the same PR, but I notice this one is already `+6,521 −3,008` (Of course, mostly in lock+requirements files.) What do you reckon @tomchop?

tomchop:
Sounds good, let's do that - I wanted to keep the number of changed files low for this one

.pylintrc
```diff
@@ -1,38 +1,70 @@
-# Pylint 2.1.x - 2.2.x configuration file
+# Pylint 2.10.x configuration file
 #
 # This file is generated by l2tdevtools update-dependencies.py, any dependency
 # related changes should be made in dependencies.ini.
-[MAIN]
+[MASTER]
```
sa3eed3ed, line 5: is this changed to master because you generated the file in a branch and that is the default? shouldn't this be the MAIN branch here?

tomchop: This is from the .pylintrc file config, don't know why it was changed to MAIN before: https://www.codeac.io/documentation/pylint-configuration.html

PR #644, opened 2022-07-26
Handle curses error when window is too short

ramo-j:
In curses mode, if the terminal is too short (read: has too few rows) the cursor moves off the screen, and an uncaught exception is thrown. This PR catches that exception and advises the user they may wish to make the terminal larger, and keeps running the recipe.


itsmvd (approved):
LGTM. This is not the ideal solution for this problem, but it's at least better than curses crashing inexplicably .

PR #645, opened 2022-07-28
Raw disk access

tomchop:


data/recipes/grr_artifact_grep.json
```diff
@@ -32,7 +32,7 @@
         ["keywords", "Pipe-separated list of keywords to search for (e.g. key1|key2|key3.", null],
         ["--artifacts", "Comma-separated list of artifacts to fetch (override default artifacts).", null],
         ["--extra_artifacts", "Comma-separated list of artifacts to append to the default artifact list.", null],
-        ["--use_tsk", "Use TSK to fetch artifacts.", false],
+        ["--use_raw_filesystem_access", "Use raw disk access to fetch artifacts.", true],
```
ramo-j, line 35: Changing the recipe defaults of raw access from false to true results in different artefact layouts, if memory serves. We're not worried about changing the default?

ramo-j (approved):
One small question, but otherwise, LGTM.

data/recipes/grr_artifact_grep.json
tomchop: I think the artifacts that we collect remain the same regardless of the access method (if I understood your question correctly)

ramo-j: Perhaps I'm misremembering then, I thought the grr zip archive output was structured differently.

ramo-j (approved):

PR #654, opened 2022-08-10
Remove race condition on tar file creation in utils.py

### LINKED ISSUE #653: Race condition in generating a compressed filename.
There is a race condition in filename generation for compressed files.  

The call to ```int(time()``` is of second precision which can trigger the output file exists error.

https://github.com/log2timeline/dftimewolf/blob/main/dftimewolf/lib/utils.py#L34

ramo-j:
Closes #653

### DIFF
--- a/dftimewolf/lib/utils.py
+++ b/dftimewolf/lib/utils.py
@@ -3,10 +3,11 @@
 
 import argparse
 import os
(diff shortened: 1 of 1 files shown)


sydp (approved):
LGTM

PR #666, opened 2022-09-19
Fix links to recipes in doc

helferd:
The section in the documentation describing how to create recipes contained two links that didn't work. These habe been fixed. However, recipe `local_plaso` mentioned in the text no longer exists. It has been replaced by the recipe `plaso_ts` as an example.


tomchop (approved):

tomchop:
Thanks for the fix!

PR #673, opened 2022-11-18
GRRYaraScanner module

tomchop:
This PR introduces a `GRRYaraScanner` module. It loads `containers.YaraRule` containers and launches a `YaraProcessScan` GRR flow against a given set of hosts. The output is renderd and stored as a pandas DataFrame enumerating host, process, and match information.

### DIFF
--- a/tests/lib/containers/containers.py
+++ b/tests/lib/containers/containers.py
@@ -6,6 +6,42 @@ import unittest
 import pandas
 from dftimewolf.lib.containers import containers
 
+CONTAINER_CLASSES = [
+  containers.AWSS3Object,
+  containers.AWSSnapshot,
+  containers.AWSVolume,
+  containers.DataFrame,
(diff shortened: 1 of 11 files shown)


tests/lib/containers/containers.py
```diff
@@ -6,6 +6,42 @@
 import pandas
 from dftimewolf.lib.containers import containers
 
+CONTAINER_CLASSES = [
```
ramo-j, line 9: [Can save for a future PR] Could we dynamically generate this list somehow? So when new containers are created, the author doesn't need to know it should be added here.

ramo-j (approved):
LGTM

tests/lib/containers/containers.py
tomchop: Opened https://github.com/log2timeline/dftimewolf/issues/674

PR #676, opened 2022-11-25
New `YetiYaraCollector` module + `grr_yarascan` recipe

tomchop:
* Introduces a new `YetiYaraCollector` module, which fetches a list of Yara rules from a Yeti instance ([repo](https://github.com/yeti-platform/TibetanBrownBear)).
* Introduces a new recipe, `grr_yarascan` which leverages `YetiYaraCollector` to collect Yara rules, and `GRRYaraScanner` to scann a specific host with them.

### DIFF
--- a/dftimewolf/cli/dftimewolf_recipes.py
+++ b/dftimewolf/cli/dftimewolf_recipes.py
@@ -71,7 +71,8 @@ MODULES = {
   'VTCollector' : 'dftimewolf.lib.collectors.virustotal',
   'WorkspaceAuditCollector': 'dftimewolf.lib.collectors.workspace_audit',
   'WorkspaceAuditTimesketch': 'dftimewolf.lib.processors.workspace_audit_timesketch',
-  'GCPCloudResourceTree': 'dftimewolf.lib.processors.gcp_cloud_resource_tree'
+  'GCPCloudResourceTree': 'dftimewolf.lib.processors.gcp_cloud_resource_tree',
+  'YetiYaraCollector': 'dftimewolf.lib.collectors.yara'
 
 }
 # pylint: enable=line-too-long
--- a/dftimewolf/cli/dftimewolf_recipes.py
+++ b/dftimewolf/cli/dftimewolf_recipes.py
@@ -56,6 +56,7 @@ MODULES = {
   'GRRHuntOsqueryDownloader': 'dftimewolf.lib.collectors.grr_hunt',
   'GRROsqueryCollector': 'dftimewolf.lib.collectors.grr_hosts',
   'GRRTimelineCollector': 'dftimewolf.lib.collectors.grr_hosts',
+  'GRRYaraScanner': 'dftimewolf.lib.collectors.grr_hosts',
   'LocalFilesystemCopy': 'dftimewolf.lib.exporters.local_filesystem',
   'LocalPlasoProcessor': 'dftimewolf.lib.processors.localplaso',
   'OsqueryCollector': 'dftimewolf.lib.collectors.osquery',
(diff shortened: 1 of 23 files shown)


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -71,7 +71,8 @@
   'VTCollector' : 'dftimewolf.lib.collectors.virustotal',
   'WorkspaceAuditCollector': 'dftimewolf.lib.collectors.workspace_audit',
   'WorkspaceAuditTimesketch': 'dftimewolf.lib.processors.workspace_audit_timesketch',
-  'GCPCloudResourceTree': 'dftimewolf.lib.processors.gcp_cloud_resource_tree'
+  'GCPCloudResourceTree': 'dftimewolf.lib.processors.gcp_cloud_resource_tree',
+  'YetiYaraCollector': 'dftimewolf.lib.collectors.yara'
```
ramo-j, line 75: Nit: Can we have these alphabetically sorted to appease my OCD?

dftimewolf/lib/collectors/yara.py
```diff
@@ -0,0 +1,85 @@
...
+      api_key: The Yeti API key to use.
+      api_root: The Yeti HTTP API root, e.g. http://localhost:8080/api/
+    """
+    self.logger.info(f'Name filter: {rule_name_filter}')
+    self.rule_name_filter = rule_name_filter or ''
+    self.api_key = api_key
+    self.api_root = api_root
+
+  def Process(self) -> None:
+
```
ramo-j, line 50: Docstring missing. Side note, I feel like the linter should have picked this up?

dftimewolf/lib/collectors/yara.py
```diff
@@ -0,0 +1,85 @@
...
+    for rule in response_json:
+      intel[rule["id"]] = rule
+    self.logger.info(f'Collected {len(intel)} Yara rules from Yeti')
+
+    for rule in intel.values():
+      container = containers.YaraRule(
+          name=rule['name'], rule_text=rule['pattern'])
+      self.state.StoreContainer(container)
+
+  def PreProcess(self) -> None:
```
ramo-j, line 77: You've inherited from `BaseModule` which doesn't require these two methods be overridden (as opposed to `ThreadAwareModule`) so they could just be removed rather than exist and be empty.

dftimewolf/lib/collectors/yara.py
```diff
@@ -0,0 +1,85 @@
...
+    self.api_root = self.api_root.strip('/')
+    response = requests.post(
+        f'{self.api_root}/indicators/filter/',
+        json={'name': self.rule_name_filter, 'type': 'x-yara'},
+        headers={'X-Yeti-API': self.api_key},
+    )
+
+    response_json = response.json()
+    if response.status_code != 200:
+      self.logger.error(
```
ramo-j, line 61: Would it be better to have `self.ModuleError()` here? Noting the recipe, this module would deliver yara rules to `GRRYaraScanner` which would still run (albeit, with no rules so would bail out.)

tests/lib/collectors/yara.py
```diff
@@ -0,0 +1,79 @@
...
+                '    $s = "r4nd0m"\n'
+                '  condition:\n'
+                '    any of them\n'
+                '}',
+    'spec_version': '2.1',
+    'type': 'x-yara',
+    'valid_from': '2022-11-17T23:00:00Z'
+}]
+
+class YetiYaraCollectorTest(unittest.TestCase):
```
ramo-j, line 33: I reckon it might be worth adding in a failure case, that is, if the server response is something other than `200 OK`.

ramo-j (changes requested):

dftimewolf/lib/collectors/yara.py
tomchop: My thinking was that if we have other modules in a recipe that ship Yara sigs, we might still want to go through with those. But I'm happy to make this a critical error if you feel strongly about this.

dftimewolf/lib/collectors/yara.py
tomchop: Weird, mypy gave me trouble when I did this before. OK, let's see what happens if I remove it...

dftimewolf/lib/collectors/yara.py
ramo-j: Makes sense.

ramo-j (approved):
LGTM

PR #678, opened 2022-11-28
Use magic in the recommended way

tomchop:
https://pypi.org/project/python-magic/ recommends doing`magic.from_file` instead of instantiating a Magic object.


sa3eed3ed (approved):
It seem like using the ` f = magic.Magic(uncompress=True)` way makes problem when running across multiple threads, https://github.com/ahupp/python-magic:
```
it's not safe for sharing across multiple threads
```

Issue #679, opened 2022-12-02
gcp_logging_collect recipe fails silently when it hits Cloud Logging API limits


bskeggs:
When running the gcp_logging_collect recipe, the collection will fail silently if it exceeds the cloud logging API quota limit. 

Based on my testing, this will first issue a 'google_api_exceptions.TooManyRequests' exception. Following this, all subsequent queries from the same logging client will be returned with a 'google_api_exceptions.StopIteration' exception. This occurs even when I implemented a 2 minute delay between requests after the first 'TooManyRequests' exception.

Due to this, I believe the previous logging retries fix from https://github.com/log2timeline/dftimewolf/commit/32e39780d808cc506c19580aea68e3b1c1810ede may not achieve any successful retries, as a new logging client is not generated after hitting the API limit.

ramo-j:
Obsolete.

PR #680, opened 2022-12-02
Feature backoff and delay

bskeggs:
This PR adds two optional arguments to the gcp_logging_collect recipe: backoff and delay. These are introduced to help manage GCP's logging api limits, which I have regularly exceeded when using dftimewolf.

Both are disabled by default, so should not impact existing dftimewolf workflows.

Backoff is a boolean value, which if True, will retry log collection if the GCP logging API quota is exceeded. Due to the way the logging API handles subsequent requests after exceeding an the API limit (https://github.com/log2timeline/dftimewolf/issues/679), a new cloud logging client must be created for the re-try. This means the collection must start again. If no delay was configured, the collection will be retried with a delay of 1s between each request. If a delay was configured, the collection will be retried with a delay of 2x the set value.

Delay is the number of seconds to delay between each API request.

When dftimewolf if run without delay or backoff arguments, and the API limit is exceeded, the output will look like this:
```
poetry run dftimewolf gcp_logging_collect 'projectname' 'timestamp >= "2022-11-30T00:00:00Z" AND timestamp <= "2022-12-08T08:00:00Z"'
  [ dftimewolf       ] Debug log: /tmp/dftimewolf-run-20221202_081136_l_1y_kd9.log
  [ GCPLogsCollector ] backoff False
  [ GCPLogsCollector ] delay 0.0
  [ GCPLogsCollector ] Hit quota limit requesting GCP logs.
  [ GCPLogsCollector ] Exponential backoff was not enabled, so query has exited.
  [ GCPLogsCollector ] The collection is most likely incomplete.
  [ GCPLogsCollector ] Downloaded logs to /tmp/tmpjfwq0jhv.jsonl
```

When backoff and delay are specified, the output will look like this:
```
poetry run dftimewolf gcp_logging_collect 'projectname' 'timestamp >= "2022-11-30T00:00:00Z" AND timestamp <= "2022-12-08T08:00:00Z"' --backoff --delay 1
Messages
  [ dftimewolf       ] Debug log: /tmp/dftimewolf-run-20221202_081057_1ph_z9ws.log
  [ GCPLogsCollector ] backoff True
  [ GCPLogsCollector ] delay 1.0
  [ GCPLogsCollector ] Hit quota limit requesting GCP logs.
  [ GCPLogsCollector ] Retrying in 60 seconds with a slower query rate.
  [ GCPLogsCollector ] Due to the GCP logging API, the query must restart from the beginning
  [ GCPLogsCollector ] Setting up new logging client.
  [ GCPLogsCollector ] Restarting query with an API request rate of 1 per 2s
  [ GCPLogsCollector ] Downloaded logs to /tmp/tmpfr1nz782.jsonl
```

Please note that I had to break up the bulk of the code in the 'Process' function into distinct functions, however most of the code base is the same. I had to break it up because I needed a way to re-call the functions to generate a new logging client, and set a new output path, once the api limit was exceeded.

### DIFF
--- a/data/recipes/gcp_logging_cloudaudit_ts.json
+++ b/data/recipes/gcp_logging_cloudaudit_ts.json
@@ -14,7 +14,7 @@
         "name": "GCPLogsCollector",
         "args": {
             "project_name": "@project_name",
-            "filter_expression": "@filter_expression",
+            "filter_expression": "logName=projects/@project_name/logs/cloudaudit.googleapis.com%2Factivity timestamp>\"@start_date\" timestamp<\"@end_date\"",
             "backoff": "@backoff",
             "delay": "@delay"
         }
@@ -40,6 +40,8 @@
         ["--incident_id", "Incident ID (used for Timesketch description).", null],
         ["--sketch_id", "Timesketch sketch to which the timeline should be added.", null],
         ["--token_password", "Optional custom password to decrypt Timesketch credential file with.", ""],
-        ["--wait_for_timelines", "Whether to wait for Timesketch to finish processing all timelines.", true]
+        ["--wait_for_timelines", "Whether to wait for Timesketch to finish processing all timelines.", true],
+        ["--backoff", "If GCP Cloud Logging API query limits are exceeded, retry with an increased delay between each query to try complete the query at a slower rate.", false],
+        ["--delay", "Number of seconds to wait between each GCP Cloud Logging query to avoid hitting API query limits", 0]
     ]
 }
--- a/data/recipes/gcp_logging_cloudaudit_ts.json
+++ b/data/recipes/gcp_logging_cloudaudit_ts.json
@@ -14,7 +14,9 @@
         "name": "GCPLogsCollector",
         "args": {
             "project_name": "@project_name",
-            "filter_expression": "logName=projects/@project_name/logs/cloudaudit.googleapis.com%2Factivity timestamp>\"@start_date\" timestamp<\"@end_date\""
+            "filter_expression": "@filter_expression",
+            "backoff": "@backoff",
+            "delay": "@delay"
         }
     }, {
         "wants": ["GCPLogsCollector"],
--- a/data/recipes/gcp_logging_collect.json
+++ b/data/recipes/gcp_logging_collect.json
@@ -14,11 +14,15 @@
         "name": "GCPLogsCollector",
         "args": {
             "project_name": "@project_name",
-            "filter_expression": "@filter_expression"
+            "filter_expression": "@filter_expression",
+            "backoff": "@backoff",
+            "delay": "@delay"
         }
     }],
     "args": [
         ["project_name", "Name of the GCP project to collect logs from.", null],
-        ["filter_expression", "Filter expression to use to query GCP logs. See https://cloud.google.com/logging/docs/view/query-library for examples.", "resource.type = 'gce_instance'"]
+        ["filter_expression", "Filter expression to use to query GCP logs. See https://cloud.google.com/logging/docs/view/query-library for examples.", "resource.type = 'gce_instance'"],
+        ["--backoff", "If API query limits are exceeded, retry with an increased delay between each query to try complete the query at a slower rate.", false],
+        ["--delay", "Number of seconds to wait between each query to avoid hitting API query limits", 0]
     ]
 }
--- a/data/recipes/gcp_logging_collect.json
+++ b/data/recipes/gcp_logging_collect.json
@@ -22,7 +22,7 @@
     "args": [
         ["project_name", "Name of the GCP project to collect logs from.", null],
         ["filter_expression", "Filter expression to use to query GCP logs. See https://cloud.google.com/logging/docs/view/query-library for examples.", "resource.type = 'gce_instance'"],
-        ["--backoff", "If API query limits are exceeded, retry with an increased delay between each query to try complete the query at a slower rate.", false],
-        ["--delay", "Number of seconds to wait between each query to avoid hitting API query limits", 0]
+        ["--backoff", "If GCP Cloud Logging API query limits are exceeded, retry with an increased delay between each query to try complete the query at a slower rate.", false],
+        ["--delay", "Number of seconds to wait between each GCP Cloud Logging query to avoid hitting API query limits", 0]
     ]
 }
--- a/dftimewolf/lib/collectors/gcp_logging.py
+++ b/dftimewolf/lib/collectors/gcp_logging.py
@@ -84,6 +84,9 @@ class GCPLogsCollector(module.BaseModule):
       backoff_multiplier (str): Query delay multiplier if the API quota is met and backoff is enabled
       output_file (str): Output file name
       output_path (str): Output file path
+    
+    Returns:
+      output_path (str): Log output path (may have been updated if API quota was exceeded)
     """
     while True:
       try:
--- a/dftimewolf/lib/collectors/gcp_logging.py
+++ b/dftimewolf/lib/collectors/gcp_logging.py
@@ -3,7 +3,7 @@
 import json
 import tempfile
 import time
-from typing import Optional, Dict, Any
+from typing import Optional, Dict, Any, Tuple
 
 from google.api_core import exceptions as google_api_exceptions
 from google.auth import exceptions as google_auth_exceptions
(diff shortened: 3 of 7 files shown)


dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -42,75 +40,121 @@ def __init__(self,
...
+          self.PublishMessage('Due to the GCP logging API, the query must restart from the beginning')
+          time.sleep(60)
+          if self._delay == 0:
+            self._delay = 1
+          else:
+            self._delay *= backoff_multiplier
+          self.PublishMessage('Setting up new logging client.')
+          logging_client = self.SetupLoggingClient()
+          pages = self.ListPages(logging_client)
+          self.PublishMessage('Restarting query with an API request rate of 1 per {0}s'.format(self._delay))
```
ramo-j, line 108: Nit, here and elsewhere - we're trying to standardise on fstrings.
```suggestion
          self.PublishMessage(f'Restarting query with an API request rate of 1 per {self._delay}s')
```

dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -42,75 +40,121 @@ def __init__(self,
...
     self._project_name = ''
+    self._backoff = False
+    self._delay = 0
+
+  def OutputFile(self):
+    """Generate an output file name and path"""
+    output_file = tempfile.NamedTemporaryFile(
+        mode='w', delete=False, encoding='utf-8', suffix='.jsonl')
+    output_path = output_file.name
+    self.logger.info('Downloading logs to {0:s}'.format(output_path))
```
ramo-j, line 51:
```suggestion
    self.logger.info(f'Downloading logs to {output_path}')
```

dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -42,75 +40,121 @@ def __init__(self,
...
-          log_dict = entry.to_api_repr()
-          output_file.write(json.dumps(log_dict))
-          output_file.write('\n')
+      'Iterate through query result pages and save json logs to disk'
+      output_path = self.ProcessPages(pages, 2, output_file, output_path)
 
     except google_api_exceptions.NotFound as exception:
       self.ModuleError(
-          f'Error accessing project: {exception!s}', critical=True)
+          'Error accessing project: {0!s}'.format(exception), critical=True)
```
ramo-j, line 157: Similar to above regarding fstrings, but perhaps I'm missing a reason you changed this line?

dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -42,75 +40,121 @@ def __init__(self,
...
+        self.PublishMessage('Hit quota limit requesting GCP logs.')
+        if self._backoff is True:
+          self.PublishMessage('Retrying in 60 seconds with a slower query rate.')
+          self.PublishMessage('Due to the GCP logging API, the query must restart from the beginning')
+          time.sleep(60)
+          if self._delay == 0:
+            self._delay = 1
+          else:
+            self._delay *= backoff_multiplier
+          self.PublishMessage('Setting up new logging client.')
```
ramo-j, line 105: I'd caution against overuse of `self.PublishMessage` for lines that could instead go to `self.logger.<LEVEL>`. PublishMessage sends lines to the curses display and is meant for advising a user of general progress. This line is more relevant to debugging I feel, and so should be sent to the logger instead. (Note, anything to PublishMessage should also go to the logger.)

dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -42,75 +40,121 @@ def __init__(self,
...
+          logging_client = self.SetupLoggingClient()
+          pages = self.ListPages(logging_client)
+          self.PublishMessage('Restarting query with an API request rate of 1 per {0}s'.format(self._delay))
+          output_file, output_path = self.OutputFile()
+        else:
+          self.PublishMessage('Exponential backoff was not enabled, so query has exited.')
+          self.PublishMessage('The collection is most likely incomplete.')
```
ramo-j, line 112: Perhaps:
```suggestion
          self.PublishMessage('The collection is most likely incomplete.', is_error=True)
```

dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -42,75 +40,121 @@ def __init__(self,
...
       project_name (str): name of the project to fetch logs from.
       filter_expression (str): GCP advanced logs filter expression.
+      backoff (bool): Retry queries with an increased delay when API quotas are exceeded.
+      delay (str): Seconds to wait between retreiving results pages to avoid exceeding API quotas
     """
     self._project_name = project_name
     self._filter_expression = filter_expression
+    self._backoff = backoff
+    self._delay = float(delay)
+    self.PublishMessage(f'backoff {self._backoff}')
```
ramo-j, line 137: Similar to above on `PublishMessage` - I don't think this output is necessary, user should know what values they just put in, and for debugging and module params are included in debug output already.

ramo-j (changes requested):
Thanks for the PR! Logic looks solid to me, just a few nits, and noting the failing automated checks. 

Also noting that we don't have unit tests for this module, so raised https://github.com/log2timeline/dftimewolf/issues/681.

dftimewolf/lib/collectors/gcp_logging.py
bskeggs: There was no reason that I recall. I've changed this back - thanks

dftimewolf/lib/collectors/gcp_logging.py
bskeggs: Agree - Changed to self.logger.debug

dftimewolf/lib/collectors/gcp_logging.py
bskeggs: updated with your suggested change, thanks

dftimewolf/lib/collectors/gcp_logging.py
bskeggs: I ended up removing this message completely. After initially changing it to debug logging, I realised this is redundant as dftimewolf logs the arguments (backoff and delay) configuration to debug on execution.

bskeggs:
Thanks for the review @ramo-j . I've addressed your suggestions and push the changes. Please let me know if there is anything else I can do regarding the automated checks.

bskeggs:
Should be good to go on the linting tests now @ramo-j , thanks!

data/recipes/gcp_logging_cloudaudit_ts.json
```diff
@@ -14,7 +14,9 @@
         "name": "GCPLogsCollector",
         "args": {
             "project_name": "@project_name",
-            "filter_expression": "logName=projects/@project_name/logs/cloudaudit.googleapis.com%2Factivity timestamp>\"@start_date\" timestamp<\"@end_date\""
+            "filter_expression": "@filter_expression",
```
ramo-j, line 17: Here and in other recipes: Any arguments to be interpolated (that is, `@arg_name`) into a recipe need to be added into the `"args"` section. In this recipe, that would be `filter_expression`, `backoff` and `delay`.

Also, do we want to change up the filter expression here anyway? I didn't author the relevant module, so I'm not sure why the original `filter_expression` takes the form it does, but I assume there must have been a reason. How does changing this affect log collection? I notice `data/recipes/gcp_logging_collect.json` has the form you are changing to, so I assume it's probably fine, but we should make sure we're not breaking any expected behaviour.

data/recipes/gcp_logging_collect.json
```diff
@@ -14,11 +14,15 @@
...
+            "backoff": "@backoff",
+            "delay": "@delay"
         }
     }],
     "args": [
         ["project_name", "Name of the GCP project to collect logs from.", null],
-        ["filter_expression", "Filter expression to use to query GCP logs. See https://cloud.google.com/logging/docs/view/query-library for examples.", "resource.type = 'gce_instance'"]
+        ["filter_expression", "Filter expression to use to query GCP logs. See https://cloud.google.com/logging/docs/view/query-library for examples.", "resource.type = 'gce_instance'"],
+        ["--backoff", "If API query limits are exceeded, retry with an increased delay between each query to try complete the query at a slower rate.", false],
+        ["--delay", "Number of seconds to wait between each query to avoid hitting API query limits", 0]
```
ramo-j, line 26: As per my other comment, you'll just need to copy these additions into the other recipes you've changed.

ramo-j (approved):

PR #682, opened 2022-12-06
Bugfixes / improvements in grr_hosts

tomchop:
* Raise real error when GRR returns an error
* Deal with crashed flows
* Allow fetching of children flows


jkppr (approved):
LGTM

PR #690, opened 2022-12-20
[WIP] Move osquery flow creation to PreProcess()

sydp:
This PR changes the GRROsqueryCollector to thread on osquery flows (via a new OsqueryGrrFlow) to improve throughput when multiple queries are submitted to the module.

### DIFF
--- a/dftimewolf/lib/collectors/grr_hosts.py
+++ b/dftimewolf/lib/collectors/grr_hosts.py
@@ -1046,13 +1046,13 @@ class GRROsqueryCollector(GRRFlow):
               ignore_stderr_errors=self.ignore_stderr_errors)
 
           try:
-            flow_id = self._LaunchFlow(client, 'OsqueryFlow', hunt_args)  
+            flow_id = self._LaunchFlow(client, 'OsqueryFlow', hunt_args)
           except DFTimewolfError as error:
             self.ModuleError(
                 f'Error raised while launching flow: {error.message}')
             continue
-          
-          self.state.StoreContainer(containers.GRRFlow(
+
+          self.state.StoreContainer(containers.GrrFlow(
                 hostname=host_container.hostname,
                 flow_id=flow_id))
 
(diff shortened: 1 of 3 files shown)


ramo-j:
I'm not sure moving launching of flows into PreProcess is the right way to go. Generally, Preprocess was intended for preparation steps that need to be taken only once, so it only runs once. That means with this change, all the flow launches would happen in sequence which seems to be counter to what you wanted (ie, `thread on osquery flows (via a new OsqueryGrrFlow) to improve throughput`). Granted, launching a flow is not a slow operation, and this might only be an issue if you want to query 100s of hosts.

I would suggest that PreProcess is used to organise the hosts and queries into the `OsqueryGrrFlow` containers, then within Process launch and wait for completion of a single flow based in the `OsqueryGrrFlow` container.

sydp:
> I'm not sure moving launching of flows into PreProcess is the right way to go. Generally, Preprocess was intended for preparation steps that need to be taken only once, so it only runs once. That means with this change, all the flow launches would happen in sequence which seems to be counter to what you wanted (ie, `thread on osquery flows (via a new OsqueryGrrFlow) to improve throughput`). Granted, launching a flow is not a slow operation, and this might only be an issue if you want to query 100s of hosts.
> 

My understanding is that launching is not a blocking operation so not sure if multithreading on this would yield much benefit right?  

Also I think in general if one was to target 100s of hosts a hunt would be a more suitable approach.  

> I would suggest that PreProcess is used to organise the hosts and queries into the `OsqueryGrrFlow` containers, then within Process launch and wait for completion of a single flow based in the `OsqueryGrrFlow` container.

Ack, I'll have another think about how to reimplement this.

tomchop:
Alternatively, we could also keep the threading as it is (on hosts) but _also_ thread the launching of the flows "manually" (e.g. using ThreadPoolExecutor) within the module itself.

sydp:
Closing in favour of https://github.com/log2timeline/dftimewolf/pull/696

Issue #681, opened 2022-12-04
Unit tests for dftimewolf/lib/collectors/gcp_logging.py


ramo-j:

ramo-j:
Obsolete.

PR #693, opened 2022-12-22
Change gcp_logging* recipes to use Plaso instead of GCPLoggingTimesketch

### LINKED ISSUE #691: Change gcp_logging* recipes to use Plaso instead of GCPLoggingTimesketch
Plaso now has a [parser for GCP Logging](https://github.com/log2timeline/plaso/blob/main/plaso/parsers/jsonl_plugins/gcp_log.py) which parses more fields from logs than the current dedicated module for parsing GCP logs [GCPLoggingTimesketch](https://github.com/log2timeline/dftimewolf/blob/main/dftimewolf/lib/processors/gcp_logging_timesketch.py).

Recipes that work on GCP logs should be adjusted to use Plaso (local or via Turbinia) for parsing similar to the `aws_logging_ts` recipe.

The GCPLoggingCollector module will have to be changed to output File containers rather the GCPLogs containers.

jonathan-greig:
Fixes #691


dftimewolf/lib/collectors/gcp_logging.py
```diff
@@ -139,9 +139,7 @@ def Process(self) -> None:
     self.PublishMessage(f'Downloaded logs to {output_path}')
     output_file.close()
 
-    logs_report = containers.GCPLogs(
-        path=output_path, filter_expression=self._filter_expression,
-        project_name=self._project_name)
+    logs_report = containers.File(self._filter_expression, output_path)
```
ramo-j, line 142: Noting we don't have unit tests for this module. #681 has already been raised.

ramo-j (approved):

PR #696, opened 2022-12-22
Update GRROsqueryCollector to use threadpoolexecutor

sydp:
As suggested, using threadpoolexecutor to run osquery in multiple threads.


dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +981,85 @@ def _DownloadResults(self,
...
 
     Raises:
       DFTimewolfError: if no artifacts specified nor resolved by platform.
     """
-    for client in self._FindClients([container.hostname]):
-      osquery_containers = self.state.GetContainers(containers.OsqueryQuery)
+    osquery_containers = self.state.GetContainers(containers.OsqueryQuery)
 
+    host_osquery_futures = []
+    with ThreadPoolExecutor() as executor:
```
tomchop, line 1053: What's the default max threads here? Should we limit it to e.g. 5 if we're concerned about not running too many flows concurrently on machines?

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +981,85 @@ def _DownloadResults(self,
...
-
-          self.state.StoreContainer(dataframe_container)
+        host_osquery_future = executor.submit(
+          self._ProcessQuery, container.hostname, osquery_container)
+        host_osquery_futures.append(host_osquery_future)
+
+    for host_osquery_future in host_osquery_futures:
+      if host_osquery_future.exception():
+        raise DFTimewolfError('Error with osquery future') \
+            from host_osquery_future.exception()
```
tomchop, line 1062: Please use `self.ModuleError` or `self.logger.error/warning` depending on how you want the execution flow of your recipe to go.  Also, let's try to keep code-related jargon (e.g. `future`) out of the error messages which are for the user.

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +981,85 @@ def _DownloadResults(self,
...
+        timeout_millis=self.timeout_millis,
+        ignore_stderr_errors=self.ignore_stderr_errors)
+
+    try:
+      flow_id = self._LaunchFlow(client, 'OsqueryFlow', hunt_args)
+      self._AwaitFlow(client, flow_id)
+    except DFTimewolfError as error:
+      self.ModuleError(
+          f'Error raised while launching/awaiting flow: {error.message}')
+      return
```
tomchop, line 1008: it would be interesting to also account for reporting when a flow gets canceled because of a timeout (given that we provide the timeout threshold in this function). (I suspect this would mess with the rest of your code here)

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +981,85 @@ def _DownloadResults(self,
...
+      hostname: str,
+      osquery_container: containers.OsqueryQuery
+  ) -> None:
+    """Processes an osquery flow on a GRR client.
+
+    Args:
+      hostname (str): the GRR client hostname.
+      osquery_container (containers.OsqueryQuery): the osquery.
+    """
+    client = self._GetClientBySelector(hostname)
```
tomchop, line 995: Can we extract this call outside the threadpoolexecutor map call? Having it here means that we fetch the GRR client every time we want to launch a flow, and we only need it once (we can pass a GRR client object here instead of a `hostname`)

tomchop (changes requested):

dftimewolf/lib/collectors/grr_hosts.py
tomchop: So running this on my machine (8 cores), means that we'll be scheduling 40 OSQueryFlows on the target machines. Maybe that's a tad too much to leave a default value?

sydp: FWIW, as of 3.8 max_workers is now ```min(32, os.cpu_count() + 4)```.

I've set this arbitrarily to 4 for now but I also believe GRR will throttle the flows anyway so it doesn't overflow the client.

dftimewolf/lib/collectors/grr_hosts.py
sydp: Done

dftimewolf/lib/collectors/grr_hosts.py
sydp: Agree on both - it would be interesting and imo it would make the function flow more confusing (pun not intended).  For now, I would opt to handle flow timeouts more generically in the GRRFlow module (_AwaitFlow most likely).  I've created https://github.com/log2timeline/dftimewolf/issues/698 to look into this.

dftimewolf/lib/collectors/grr_hosts.py
sydp: Done

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +982,87 @@ def _DownloadResults(self,
...
+      hostname: str,
+      client: Client,
+      osquery_container: containers.OsqueryQuery
+  ) -> None:
+    """Processes an osquery flow on a GRR client.
+
+    Args:
+      hostname (str): the GRR client hostname.
+      client (Client): the Grr Client.
+      osquery_container (containers.OsqueryQuery): the osquery.
```
tomchop, line 996:
```suggestion
      hostname: the GRR client hostname.
      client: the GRR Client.
      osquery_container: the OSQuery.
```

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +982,87 @@ def _DownloadResults(self,
...
+    host_osquery_futures = []
+    with ThreadPoolExecutor(self.GetQueryThreadPoolSize()) as executor:
+      for osquery_container in osquery_containers:
+        host_osquery_future = executor.submit(
+          self._ProcessQuery, container.hostname, client, osquery_container)
+        host_osquery_futures.append(host_osquery_future)
 
-          self.state.StoreContainer(dataframe_container)
+    for host_osquery_future in host_osquery_futures:
+      if host_osquery_future.exception():
```
tomchop, line 1063: How will this behave with `self.ModuleError`? Maybe this was your idea from the start, to use DFTimewolfExceptions in `_ProcessQuery` so that they would get caught here?

tomchop: I would also add a check that if ALL threads errored out, then we do a critical `ModuleError` as there's probably no point in continuing the recipe.

dftimewolf/lib/collectors/grr_hosts.py
```diff
@@ -980,64 +982,87 @@ def _DownloadResults(self,
...
+    if not results:
+      results_container = containers.OsqueryResult(
+          name=name,
+          description=description,
+          query=query,
+          hostname=hostname,
+          data_frame=pd.DataFrame(),
+          flow_identifier=flow_identifier,
+          client_identifier=client_identifier)
+      self.state.StoreContainer(results_container)
```
tomchop, line 1027: What's the point in storing an empty dataframe here? Wouldn't it be better just to not store any container?

tomchop (changes requested):

dftimewolf/lib/collectors/grr_hosts.py
sydp: I have it as an empty dataframe as the corresponding container attribute is currently not optional.  It also simplifies the logic in downstream processing of the container.

dftimewolf/lib/collectors/grr_hosts.py
sydp: > How will this behave with self.ModuleError? Maybe this was your idea from the start, to use DFTimewolfExceptions in _ProcessQuery so that they would get caught here?

This should catch any "critical" ModuleErrors as well as any other unhandled exceptions raised in the worker.   I don't think it was my original intention to do that here - it was more to handle any unknown thread related issues that might arise during execution.

> I would also add a check that if ALL threads errored out, then we do a critical ModuleError as there's probably no point in continuing the recipe.

I'm assuming you mean all threads for a single host in which case I think it would be OK to continue the other threads processing other hosts?   Main scenario that comes to mind when all threads would fail would be if there are timeout issues due to the host being unresponsive/offline.

dftimewolf/lib/collectors/grr_hosts.py
tomchop: My question was more "why add a container at all" if the dataframe is going to be empty anyways.

dftimewolf/lib/collectors/grr_hosts.py
tomchop: Yeah that's a fair point. Thanks for clarifying.

dftimewolf/lib/collectors/grr_hosts.py
sydp: Oops my bad, missed the second question.  My rationale doing it this way that was no result (i.e. empty data) is still a result and is useful feedback downstream to let the module/user know that the query was successful and there was no result.

tomchop: OK, that makes sense, thanks!

tomchop (approved):

Issue #702, opened 2023-01-10
Strongly type dftimewolf input parameters


ramo-j:
From OOB discussion:

```
dfTimewolf takes a lot of different input parameters (defined in recipes), but they're often the same
"type": dates, ticket IDs, etc. It's always up to the recipe / module writer to parse, validate, and
convert these into the appropriate datatype (e.g. dates), which creates a lot of repetition and opens
the codebase to more bugs.

It would be good to define an ontology of acceptable parameters and how to parse them and
migrate the existing ones.
```

PR #703, opened 2023-01-12
Adjusted recipe argument parsing

ramo-j:
Adjusted argument parsing from recipes to use a new dataclass. Change is transparent to a user, but is prep work for #702

### DIFF
--- a/dftimewolf/lib/recipes/manager.py
+++ b/dftimewolf/lib/recipes/manager.py
@@ -38,8 +38,8 @@ class RecipesManager(object):
     args = []
     for arg_list in json_dict['args']:
       arg = RecipeArgs(switch=arg_list[0],
-                             help_text=arg_list[1],
-                             default=arg_list[2])
+                       help_text=arg_list[1],
+                       default=arg_list[2])
       args.append(arg)
     del json_dict['args']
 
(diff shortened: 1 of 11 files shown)


dftimewolf/lib/recipes/manager.py
```diff
@@ -35,7 +35,12 @@ def _ReadRecipeFromFileObject(
     description = json_dict['description']
     del json_dict['description']
 
-    args = json_dict['args']
+    args = []
+    for arg_list in json_dict['args']:
+      arg = RecipeArgs(switch=arg_list[0],
+                             help_text=arg_list[1],
```
sydp, line 41: fix indentation for the next two lines

sydp (dismissed):
One minor style nit but otherwise LGTM

dftimewolf/lib/recipes/manager.py
sydp: or  you may be able to unpack from the list directly into the dataclass

ramo-j: Done. Went with just fixing the indentation, as I know there's going to be an optional element to this dataclass in a later PR, I can look into your other suggestion.

sydp (approved):
LGTM

PR #705, opened 2023-01-20
Recipe argument validation

ramo-j:
This PR implements recipe argument validation, and a small number of validators. 

The validators implemented include:

* AWS Region
* GCP Zone
* Regex
* Subnet

One recipe has a small number of validators added to show the methodology works - And to course correct if we spot problems before writing all the validators. 

Recipe parameters that do not specify a validator use a default validator which always passes. Failures raise a `RecipeArgsValidatorError` which is handled at the orchestration layer.


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -275,6 +277,27 @@ def ParseArguments(self, arguments: List[str]) -> None:
...
+      except errors.RecipeArgsValidatorError as exception:
+        error_messages.append(f'Argument validation error: "{arg.switch}" with '
+            f'value "{self.state.command_line_options[switch]}" gave error: '
+            f'{str(exception)}')
+
+    if error_messages:
+      for message in error_messages:
+        logger.critical(message)
+      raise errors.RecipeArgsValidatorError(
+          'At least one argument failed validation')
```
tomchop, line 299: I think this would be best placed directly in argparse using the `type` parameter of `add_argument`: https://docs.python.org/3/library/argparse.html#type

If your validators can be reduced to simple functions, then we can just map the string provided in the `format` key of the arg with your validation function.

Or, given that the set of parameters can be a bit complex (e.g. `regex` depends on the actual regex being passed), you could also imagine a function that takes in the last element of the tuple in `args`, and passes that to a function that "picks" which validator to run, with which parameters.

In addition to validating parameters, this would also enable us to e.g. split comma-separated values before gettingp assed to SetUp(), which would avoid us having to write this logic in every module.

WDYT?

tomchop (changes requested):

dftimewolf/cli/dftimewolf_recipes.py
ramo-j: Overall, that makes sense, I'll give it a crack and see if I can get something working. That said:

> If your validators can be reduced to simple functions, then we can just map the string provided in the format key of the arg with your validation function.

They probably can, but, from the docs: `The argument to type can be any callable that accepts a single string.` Being a single string makes it tough, since we can't get the optional parameters in that way. Even with your suggestion `you could also imagine a function that takes in the last element of the tuple in args`, we need to pass in a dict (or convert it to, and then back from, a `str`.)

> In addition to validating parameters, this would also enable us to e.g. split comma-separated values before gettingp assed to SetUp()

This would be ideal, but I suspect it might end up as a breaking change, which I was hoping to avoid - As it will mean that all modules need to be updated to accept a list where before they required a single item.

dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -422,6 +445,14 @@ def RunTool(cdm: Optional[CursesDisplayManager] = None) -> bool:
 
   tool.state.LogExecutionPlan()
 
+  try:
+    tool.ValidateArguments()
+  except errors.RecipeArgsValidatorError as exception:
+    if cdm:
+      cdm.EnqueueMessage('dftimewolf', str(exception), True)
+    logger.critical(str(exception))
+    return False
```
tomchop, line 454: I'd move this up right before LogExecutionPlan()

dftimewolf/lib/args_validator.py
```diff
@@ -0,0 +1,226 @@
...
+  def __init__(self) -> None:
+    """Initialise."""
+
+  @abc.abstractmethod
+  def Validate(
+      self, operand: Any, validator_params: Optional[Dict[str, Any]]) -> None:
+    """Validate the parameter.
+
+    Subclasses that override this should raise errors.RecipeArgsValidatorError
+    on validation failures, with a string describing the problem."""
```
tomchop, line 27: A description of the parameters here would be nice.

dftimewolf/lib/args_validator.py
```diff
@@ -0,0 +1,226 @@
...
+  _regions = [
+    'af-south-1', 'ap-east-1', 'ap-northeast-1', 'ap-northeast-2',
+    'ap-northeast-3', 'ap-south-1', 'ap-south-2', 'ap-southeast-1',
+    'ap-southeast-2', 'ap-southeast-3', 'ap-southeast-4', 'ap-southeast-6',
+    'ca-central-1', 'ca-west-1', 'cn-north-1', 'cn-northwest-1', 'eu-central-1',
+    'eu-central-2', 'eu-north-1', 'eu-south-1', 'eu-south-2', 'eu-west-1',
+    'eu-west-2', 'eu-west-3', 'il-central-1', 'me-central-1', 'me-south-1',
+    'sa-east-1', 'us-east-1', 'us-east-2', 'us-gov-east-1', 'us-gov-west-1',
+    'us-west-1', 'us-west-2'
+  ]
```
tomchop, line 91: Make this a set, so that the `in` operation is faster.

```suggestion
  _regions = {
    'af-south-1', 'ap-east-1', 'ap-northeast-1', 'ap-northeast-2',
    'ap-northeast-3', 'ap-south-1', 'ap-south-2', 'ap-southeast-1',
    'ap-southeast-2', 'ap-southeast-3', 'ap-southeast-4', 'ap-southeast-6',
    'ca-central-1', 'ca-west-1', 'cn-north-1', 'cn-northwest-1', 'eu-central-1',
    'eu-central-2', 'eu-north-1', 'eu-south-1', 'eu-south-2', 'eu-west-1',
    'eu-west-2', 'eu-west-3', 'il-central-1', 'me-central-1', 'me-south-1',
    'sa-east-1', 'us-east-1', 'us-east-2', 'us-gov-east-1', 'us-gov-west-1',
    'us-west-1', 'us-west-2'
  }
```

tomchop: (same for other similar validators)

dftimewolf/lib/args_validator.py
```diff
@@ -0,0 +1,226 @@
...
+      raise errors.RecipeArgsValidatorError('Invalid GCP zone name')
+
+
+class RegexValidator(CommaSeparatedValidator):
+  """Validates a string according to a regular expression."""
+
+  def __init__(self) -> None:
+    """Initialise."""
+    super().__init__()
+    self.name = 'regex'
```
tomchop, line 162: Why have the `name` attribute as an object attribute instead of a class attribute (like in AttribtueContainers)? Can the name attribute of a validator object change throughout its lifetime?

tomchop (changes requested):

dftimewolf/cli/dftimewolf_recipes.py
ramo-j: I've put together a comparison PR, #706 that starts down the route of using validator functions in the `type` parameter of `argparse.add_parameter`. There's a few drawbacks, including less control over error messages, losing the comma separation, and some pretty funky function signatures which IMO impacts readability + maintainablility. We'd also need a new validator instance for each validation, since now the params are stored as class members (so, we can't reuse a single Regex Validator for multiple regexes in the same recipe, for example.)

dftimewolf/cli/dftimewolf_recipes.py
ramo-j: Done.

dftimewolf/lib/args_validator.py
ramo-j: Done.

dftimewolf/lib/args_validator.py
ramo-j: Done.

dftimewolf/lib/args_validator.py
ramo-j: Changed to class attribute.

dftimewolf/cli/dftimewolf_recipes.py
tomchop: Yup, that's fair, thanks for taking time to write the alternate PR. It's much clearer the way it's done here.

dftimewolf/lib/args_validator.py
```diff
@@ -0,0 +1,294 @@
...
+
+from typing import Any, Dict, Optional
+
+from dftimewolf.lib import errors
+
+
+class AbstractValidator(abc.ABC):
+  """Base class for validator objects."""
+
+  VALIDATOR_NAME: str = None  # type: ignore
```
tomchop, line 15: nit: We can probable call these NAME, since, we're doing things like validator.VALIDATOR_NAME

tomchop (dismissed):
One minor nit, otherwise LGTM

dftimewolf/lib/args_validator.py
ramo-j: Done.

tomchop (approved):

PR #708, opened 2023-01-30
Don't run validator if argument is optional and not present

ramo-j:
Doesn't run validators if the argument is optional and not specified. Currently, for an optional argument if a validator is specified but no argument passed at runtime, the validator tries to test `None`.

Also, correct the validator for `aws_subnet` - a regex of an AWS subnet ID, not a literal subnet value.


dftimewolf/lib/args_validator.py
```diff
@@ -44,7 +44,7 @@ class CommaSeparatedValidator(AbstractValidator):
 
   def Validate(self,
                operand: str,
-               validator_params: Dict[str, Any]) -> None:
+               validator_params: Optional[Dict[str, Any]]) -> None:
```
sydp, line 47: Should this have a default ```=None``` since it is now annotated as optional?

dftimewolf/lib/args_validator.py
```diff
@@ -123,7 +123,7 @@ class AWSRegionValidator(AbstractValidator):
 
   def Validate(self,
                operand: Any,
-               validator_params: Dict[str, Any]) -> None:
+               validator_params: Optional[Dict[str, Any]]) -> None:
```
sydp, line 126: Similarly - ```=None```?

dftimewolf/lib/args_validator.py
```diff
@@ -200,7 +200,7 @@ class RegexValidator(CommaSeparatedValidator):
 
   def ValidateSingle(self,
                      operand: str,
-                     validator_params: Dict[str, Any]) -> None:
+                     validator_params: Optional[Dict[str, Any]]) -> None:
```
sydp, line 203: ```=None```?

sydp (dismissed):
Minor type annotation nit otherwise LGTM

dftimewolf/lib/args_validator.py
ramo-j: Done.

dftimewolf/lib/args_validator.py
ramo-j: Done.

dftimewolf/lib/args_validator.py
ramo-j: Done.

sydp (approved):
LGTM

PR #712, opened 2023-02-01
Added a datetime args validator

ramo-j:
Added an argument validator for dates + times. 

Also, added in parameter substitution into the `args.format` section of a recipe, so arguments can be compared, as this new DatetimeValidator does.

### DIFF
--- a/dftimewolf/lib/args_validator.py
+++ b/dftimewolf/lib/args_validator.py
@@ -4,7 +4,9 @@ import abc
 import ipaddress
 import re
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
+
+import datetime
 
 from dftimewolf.lib import errors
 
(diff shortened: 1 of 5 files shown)


dftimewolf/lib/args_validator.py
```diff
@@ -246,6 +248,112 @@ def ValidateSingle(self,
...
+
+  def _ValidateOrder(self,
+                     first: Union[str, datetime.datetime],
+                     second: Union[str, datetime.datetime],
+                     format_string: str) -> None:
+    """Validates date ordering.
+
+    Args:
+      first: The earlier date.
+      second: The later date
```
tomchop, line 337:
```suggestion
      second: The later date.
```

tests/lib/args_validator.py
```diff
@@ -194,6 +194,71 @@ def test_ValidateFailure(self):
         self.validator.Validate(value, params)
 
 
+class DatetimeValidatorTest(unittest.TestCase):
```
tomchop, line 197: Could we also add a test for the variable interpolation here? i.e. that specifying two args that depend on one another (like in your example) works / fails as intended

tomchop (approved):
One small nit and requests for a test. Otherwise, LGTM!

tests/lib/args_validator.py
ramo-j: Raised 713. I don't think the test should go here though, but where that interpolation is performed, since by the time the values make it to DatetimeValidator, there's no indication whether they were interpolated or not.

ramo-j:
Accepting your suggestion still dismisses your review? Nice one github.

tomchop:
> Accepting your suggestion still dismisses your review? Nice one github.

Yeah, GitHub isn't great about this. I guess if the LGTM is given, it's OK to "force-merge" after the changes are in.

tomchop (approved):

PR #715, opened 2023-02-03
Added more Args validators

ramo-j:
Added 3 more argument validators:

* FQDNValidator
* GRRHostValidator
* URLValidator

### DIFF
--- a/dftimewolf/lib/args_validator.py
+++ b/dftimewolf/lib/args_validator.py
@@ -355,6 +355,39 @@ class DatetimeValidator(AbstractValidator):
           f'{first} is after {second} but it should be the other way around')
 
 
+class FQDNValidator(RegexValidator):
+  """Validator for FQDNs."""
+
+  NAME = 'fqdn'
+  FQDN_REGEX = (
+      r'(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{1,63}(?<!-)\.)+[a-zA-Z]{2,63}$)')
+  # Source: https://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation#20204811  # pylint: disable=line-too-long
+  # Retrieved 2023-02-03
+
+  def Validate(self,
+               operand: str,
+               validator_params: Optional[Dict[str, Any]] = None) -> None:
+    """Validate FQDNs.
+
+    Args:
+      operand: The FQDNs to validate.
+      validator_params: May contain "comma_separated": True|False.
+
+    Raises:
+      errors.RecipeArgsValidatorError: Raised if argument fails validation.
+    """
+    if not validator_params:
+      validator_params = {}
+    validator_params['regex'] = self.FQDN_REGEX
+
+    try:
+      super().Validate(operand, validator_params)
+    except errors.RecipeArgsValidatorError:
+      # Give a nicer error message than the regex failure
+      raise errors.RecipeArgsValidatorError(
+          f"'{operand}' contains an invalid hostname.")
+
+
 class ValidatorManager:
   """Class that handles validating arguments."""
 
@@ -365,6 +398,7 @@ class ValidatorManager:
 
     self.RegisterValidator(AWSRegionValidator())
     self.RegisterValidator(DatetimeValidator())
+    self.RegisterValidator(FQDNValidator())
     self.RegisterValidator(GCPZoneValidator())
     self.RegisterValidator(RegexValidator())
     self.RegisterValidator(SubnetValidator())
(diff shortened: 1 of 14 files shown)


dftimewolf/lib/args_validator.py
```diff
@@ -355,6 +356,113 @@ def _ValidateOrder(self,
...
+
+    Raises:
+      errors.RecipeArgsValidatorError: Raised if argument fails validation.
+    """
+    if not validator_params:
+      validator_params = {}
+
+    # Grr Host IDs can be FQDNs, test that first
+    try:
+      FQDNValidator.ValidateSingle(self, operand, validator_params)
```
sydp, line 419: This calling convention is a bit odd - this looks like you are invoking a classmethod on FQDNValidator with a reference to self?

dftimewolf/lib/args_validator.py
```diff
@@ -355,6 +356,113 @@ def _ValidateOrder(self,
...
+    except errors.RecipeArgsValidatorError:
+      # Hostname validation failed, check for Grr client ID.
+      validator_params['regex'] = self.GRR_REGEX
```
sydp, line 422: Since this validator is intended to validate a GRR hostname, would it make more sense to check that first and then if that fails, clear validator_params and then call super() again to check the hostname?

That might remove the need to use the ClassName.ValidateSingle(...) calls you are doing here.

tests/lib/args_validator.py
```diff
@@ -257,6 +257,123 @@ def test_ValidateOrder(self):
...
+
+    self.validator.Validate(','.join(fqdns), {'comma_separated': True})
+
+  def test_ValidationFailure(self):
+    """Tests validation failures."""
+    fqdns = [
+        'value',
+        '10.100.0.100',  # Needs scheme
+        'http://one.*.com'
+        ]
```
sydp, line 364: i'd prob remove indentation here or bring it up to the next line

sydp (changes requested):

tests/lib/args_validator.py
```diff
@@ -257,6 +257,123 @@ def test_ValidateOrder(self):
...
 
+class FQDNValidatorTest(unittest.TestCase):
+  """Tests the FQDNValidator class."""
+
+  def setUp(self):
+    """Setup."""
+    self.validator = args_validator.FQDNValidator()
+
+  def test_Init(self):
+    """Tests initialisation."""
```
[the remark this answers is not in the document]
ramo-j: :vomiting_face:

dftimewolf/lib/args_validator.py
ramo-j: `self.Method(arg)` is the same as `ClassName.Method(self, arg)` but here I needed to specify which version of `ValidateSingle` is used, from multiple potential parent classes since `self.` or `super()` seemed to pick wrong versions while writing this.

dftimewolf/lib/args_validator.py
ramo-j: The order (Grr client ID vs FQDN first) probably doesn't make that much difference, it's arbitrary. 

Also, `ValidateSingle` is indeed the correct method to go with here, as walking up the class inheritance ladder is `CommaSeparatedValidator` which requires `ValidateSingle` rather than `Validate` be overridden.

tests/lib/args_validator.py
```diff
@@ -257,6 +257,123 @@ def test_ValidateOrder(self):
...
+    self.assertEqual(self.validator.NAME, 'fqdn')
+
+  def test_ValidateSuccess(self):
+    """Test successful validtion."""
```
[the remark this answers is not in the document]
ramo-j: Done.

tests/lib/args_validator.py
```diff
@@ -257,6 +257,123 @@ def test_ValidateOrder(self):
...
+  def setUp(self):
+    """Setup."""
+    self.validator = args_validator.GRRHostValidator()
+
+  def test_Init(self):
+    """Tests initialisation."""
+    self.assertEqual(self.validator.NAME, 'grr_host')
+
+  def test_ValidateSuccess(self):
+    """Test successful validtion."""
```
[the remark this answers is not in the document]
ramo-j: This is what I get for copy-pasting the boilerplate. Done.

tests/lib/args_validator.py
```diff
@@ -257,6 +257,123 @@ def test_ValidateOrder(self):
...
+
+class URLValidatorTest(unittest.TestCase):
+  """Tests the URLValidator class."""
+
+  def setUp(self):
+    """Setup."""
+    self.validator = args_validator.URLValidator()
+
+  def test_Init(self):
+    """Tests initialisation."""
```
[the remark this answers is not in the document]
ramo-j: Done.

tests/lib/args_validator.py
ramo-j: Done.

sydp (approved):
LGTM

PR #719, opened 2023-02-07
Fix redraw in curses mode on SIGWINCH

ramo-j:
This PR corrects the redraw behaviour when SIGWINCH is raised when curses mode is in use.


sydp (approved):
Lgtm

PR #728, opened 2023-02-15
Add Telemetry functionality

tomchop:
New telemetry functionality for dfTimewolf.

Cloud spanner is set as an optional dependency so that users don't accidentally configure Cloud telemetry if they don't want it. Opt-in on steroids.


ramo-j:
Does this make sense to be implemented as a module? Seems to me that this would make more sense as a part of the state to run in all DFTW runs, rather than a module to be added to recipes.

dftimewolf/lib/state.py
```diff
@@ -234,31 +218,24 @@ def StoreContainer(self,
...
     """
     with self._state_lock:
       container.SetMetadata(
           interface.METADATA_KEY_SOURCE_MODULE, source_module)
       self.store.setdefault(container.CONTAINER_TYPE, []).append(container)
 
-  def StoreStats(self, stats_entry: StatsEntry) -> None:
-    """Thread-safe method to store stats in the state's stats store.
+  def LogTelemetry(self, telemetry_entry: telemetry.TelemetryEntry) -> None:
+    """Thread-safe method to store telemetry in the state's telemetry store.
```
ramo-j, line 230: It probably _is_ thread safe, but with the mutex removed, probably don't need to call that out.
```suggestion
    """Method to store telemetry in the state's telemetry store.
```

dftimewolf/lib/telemetry.py
```diff
@@ -0,0 +1,150 @@
...
+  from google.cloud import spanner  # type: ignore
+  HAS_SPANNER = True
+except ImportError:
+  HAS_SPANNER = False
+
+from dftimewolf import config
+
+
+@dataclass
+class TelemetryEntry:
```
ramo-j, line 18: Can I suggest a rename? `TelemetryEntry` to me sounds like a single data point, but this is a collection of data points. Maybe `TelemetrySet` or `TelemetryCollection`? Maybe even just `TelemetryDict`

dftimewolf/lib/telemetry.py
```diff
@@ -0,0 +1,150 @@
...
+    module_type: Type of the module that generated the telemetry.
+    module_name: Name of the module that generated the telemetry. This has the
+        same value as module_type when no runtime_name has been specified for
+        the module.
+    telemetry: Dictionary of telemetry to store. Contents are arbitrary, but
+        keys must be strings.
+  """
+  module_type: str
+  module_name: str
+  telemetry: Dict[str, str]
```
ramo-j, line 31: You mention keys must be strings in the docstring, but type the value as `str` as well. Should the type be `dict[str, Any]`?

ramo-j (approved):
LGTM.

In terms of the failing pytype check, I've been working on updating pytype, and as a result the newly failing type checks. I can try and get that work merged, rather than you try and incorporate those changes here.

https://github.com/log2timeline/dftimewolf/compare/main...ramo-j:dftimewolf:typing

dftimewolf/lib/module.py
```diff
@@ -72,19 +77,21 @@ def SetupLogging(self, threaded: bool = False) -> None:
...
 
     Raises:
-      ValueError: If the keys in the stats dict are not strings.
+      ValueError: If the keys in the telemetry dict are not strings.
     """
-    if not all (isinstance(key, str) for key in stats.keys()):
-      raise ValueError("Stats keys must be strings.")
-    stastentry = state_lib.StatsEntry(type(self).__name__, self.name, stats)
-    self.state.StoreStats(stastentry)
+    if not all (isinstance(key, str) for key in data.keys()):
```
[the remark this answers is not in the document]
tomchop: My rationale for enforcing it was that I didn't want the database to be corrupted by bad datatypes. Spanner does indeed enforce this (so we'd get an exception from the spanner libs if we ever tried to store, say, an int), but one could imagine a scenario where the backend is more liberal, and if pytype happens to not to catch it, then it would pass onto the database. 

Given that we "soft enforce" it using datatypes, I think it's a good idea to make sure that we raise an exception earlier rather than later.

dftimewolf/lib/telemetry.py
tomchop: Done.

dftimewolf/lib/module.py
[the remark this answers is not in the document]
tomchop: On second though, I agree, let's remove the enforcment. Types should be enough.

ramo-j (approved):

PR #729, opened 2023-02-17
Argument Validator updates

ramo-j:
Changes argument validators to return a `tuple[bool, str]` rather than nothing, or raise an exception. This is better suited for cases of overlap, such as between fqdns, flat hostnames, urls and grr ids.

### DIFF
--- a/dftimewolf/lib/args_validator.py
+++ b/dftimewolf/lib/args_validator.py
@@ -35,6 +35,9 @@ class AbstractValidator(abc.ABC):
       A Tuple:
         boolean: True if operand passes validation, False otherwise.
         str: A message for validation failure. Only set if the boolean is false.
+    
+    Raises:
+      errors.RecipeArgsValidatorError
     """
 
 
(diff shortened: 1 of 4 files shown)


dftimewolf/lib/args_validator.py
```diff
@@ -361,78 +403,99 @@ def Validate(self,
...
+                         'should be the other way around')
+
+      if 'after' in validator_params and validator_params['after']:
+        if not self._ValidateOrder(
+            validator_params['after'], dt, validator_params["format_string"]):
+          return False, (f'{dt} is before {validator_params["after"]} but it '
+                         'should be the other way around')
+    except ValueError as exception:
+      raise errors.RecipeArgsValidatorError(
+          f'Error in order comparison: {str(exception)}')
```
tomchop, line 422: Shouldn't we be returning `'False', message` here?

tomchop (approved):
Minor comment, otherwise this looks good. Very golang of you! :D

dftimewolf/lib/args_validator.py
ramo-j: The error here is if there is a mismatch in the date format string, I think an exception is more appropriate.

PR #732, opened 2023-03-03
Move container.src_module_name into metdata

ramo-j:
Moving the source container info into container metadata means it can be filtered on.

### DIFF
--- a/dftimewolf/lib/module.py
+++ b/dftimewolf/lib/module.py
@@ -169,8 +169,10 @@ class BaseModule(object):
 
     self.logger.debug(f'{self.name} is retrieving {len(containers)} '
         f'{container_class.CONTAINER_TYPE} containers - pop == {pop}')
-    for item in containers:
-      self.logger.debug(f'  * {str(item)} - origin: {item.src_module_name}')
+    for c in containers:
+      self.logger.debug(
+          f'  * {str(c)} - origin: '
+          f'{c.metadata.get(interface.METADATA_KEY_SOURCE_MODULE, "Unknown")}')
 
     return containers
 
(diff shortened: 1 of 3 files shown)


dftimewolf/lib/state.py
```diff
@@ -399,8 +400,10 @@ def _RunModuleProcessThreaded(
         as executor:
       pop = not module.KeepThreadedContainersInState()
       for c in self.GetContainers(module.GetThreadOnContainerType(), pop):
-        logger.debug(f'Launching {module.name}.Process thread with {str(c)} '
-            f'from {c.src_module_name}')
+        logger.debug(
+            f'Launching {module.name}.Process thread with {str(c)} from '
```
sydp, line 404: Similarly, I'd opt to update c to maybe container or something a bit more obvious

sydp (approved):
Lgtm, though minor nit on choice of variable name

dftimewolf/lib/module.py
```diff
@@ -169,8 +169,10 @@ def GetContainers(self,
 
     self.logger.debug(f'{self.name} is retrieving {len(containers)} '
         f'{container_class.CONTAINER_TYPE} containers - pop == {pop}')
-    for item in containers:
-      self.logger.debug(f'  * {str(item)} - origin: {item.src_module_name}')
+    for c in containers:
```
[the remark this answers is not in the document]
ramo-j: I went with a single char variable because it either breaks up the f-string `{}` on 175 over multiple lines, or goes over the 80 character line length.

dftimewolf/lib/state.py
ramo-j: Same as above, line the fstring placeholder on 405 goes too long with a longer variable name.

PR #745, opened 2023-04-04
Update BigQueryCollector to run queries generated by previous modules

ramo-j:


dftimewolf/lib/collectors/bigquery.py
```diff
@@ -22,17 +22,14 @@ def __init__(self,
...
   # pylint: disable=arguments-differ
   def SetUp(self,
-            project_name: str,
-            query: str,
-            description: str,
-            pandas_output: bool) -> None:
+            project_name: str = '',
+            query: str = '',
+            description: str = '',
+            pandas_output: bool = False) -> None:
```
tomchop, line 32: Any reason we're using default arguments here? We removed them from all other modules, to prefer setting defaults in the recipe instead.

tomchop (approved):
One small nit / question.

dftimewolf/lib/collectors/bigquery.py
ramo-j: Fixed.

tomchop (approved):

PR #762, opened 2023-06-06
Remove the telemetry singleton

tomchop:
This changes the way Telemetry is instantiated. It used to be a singleton, now it's part of the `DFTimewolfTool` object.

This allows for libraries importing dfTimewolf to be able to execute workflwos that have a distinctive UUID (one per Tool instantiation).


jleaniz:
Looks like a few of the checks are failling. Please take a look

jleaniz (approved):
LGTM

PR #764, opened 2023-07-05
Improvements

tomchop:
* Add the ability to manually pass a UUID to the dfTimewolf tool (useful when imported as a lib)
* Account for parameters passed as lists in dftimewolf/lib/collectors/gce_disk_copy.py (if invoked programmatically and not from the CLI)
* Fix a typing issue


hacktobeer (approved):

giovannt0:
Awooooooo

PR #766, opened 2023-07-14
Usage of Docker in localplaso processor is defined at recipe-level

tomchop:
It's `true` by default since this is the canonical way of running Docker


rgayon (dismissed):

tomchop:
@rgayon Already approved, merging now that mypy is happy.

PR #774, opened 2023-08-15
Add log output with link to Turbinia WebUI on request creation.

hacktobeer:
Add a log message with a link to the Turbinia WebUI so a user can easily check the status of the generated request. The current OSDFIR deployment of Turbinia deploys the webui to the same http endpoint as the api server.


jleaniz (approved):
LGTM

tomchop (changes requested):
Use `PublishMessage` instead, and this will be highlighted in the curses interface

PR #779, opened 2023-09-21
Update poetry dependencies

jkppr:
This PR allows to use dftimewolf with the Google Cloud shell:
* Allow python version 3.9 usage
* Update pytype version to 2023.09.19


jkppr:
Closing this PR, since dftimewolf is using features that were introduced in python 3.10 and not available in python 3.9. E.g. `| for types` in https://github.com/log2timeline/dftimewolf/blob/main/dftimewolf/lib/telemetry.py#L41 . Hence, even when dftimewolf would install on python 3.9 it would not work.

PR #780, opened 2023-09-21
Support cloud shell

jkppr:
This PR allows to use dftimewolf with the Google Cloud shell:

* Allow python version 3.9 usage
* Update pytype version to 2023.09.19
* Modify some type checks to work with python3.9 by using `Optional[]` instead of the `|` operator.


tomchop:
pytype is complaining about code in setup.py which we should clean up anyway. Merging

PR #781, opened 2023-09-22
update typing checks

jkppr:
Adjust typing checks that I missed in #780 to be python 3.9 compatible.

### DIFF
--- a/dftimewolf/lib/collectors/gce_disk_copy.py
+++ b/dftimewolf/lib/collectors/gce_disk_copy.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Copies GCE Disks across projects."""
 
-from typing import List, Optional, Dict, Type
+from typing import List, Optional, Dict, Type, Union
 
 from googleapiclient.errors import HttpError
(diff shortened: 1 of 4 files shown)


dftimewolf/lib/collectors/gce_disk_copy.py
```diff
@@ -62,8 +62,8 @@ def SetUp(self,
             destination_project_name: Optional[str],
             source_project_name: str,
             destination_zone: str,
-            remote_instance_names: Optional[str | List[str]],
-            disk_names: Optional[str | List[str]],
+            remote_instance_names: Union[str, List[str]],
+            disk_names: Union[str, List[str]],
```
tomchop, line 66: Can also be `None`, so add it to the union (that's what `Optional` means)

tomchop (dismissed):
One nit, otherwise LGTM! Thanks :)

dftimewolf/lib/collectors/gce_disk_copy.py
jkppr: Good point. I updated the code.

PR #786, opened 2023-10-04
New collector: LocalYaraCollector

tomchop:
* Adds a `LocalYaraCollector`
* Tweaks the `gcp_turbinia_disk_copy_ts` recipe to include this collector so that files can be sent to Turbinia.


jkppr (approved):
Tested the changes and works great! Thanks for implementing this feature on short notice.

PR #794, opened 2023-10-24
More options for Timesketch auth (user/pass)

tomchop:
* Changes in the Timesketch exporter and enhancer (and recipes) with the option of using simple user / pass authentication


data/recipes/workspace_user_login_ts.json
```diff
@@ -39,6 +42,9 @@
     ["--filter_expression", "Filter expression to use to query Workspace logs. See https://developers.google.com/admin-sdk/reports/reference/rest/v1/activities/list",""],
     ["--incident_id", "Incident ID (used for Timesketch description).", null],
     ["--sketch_id", "Timesketch sketch to which the timeline should be added.", null, {"format": "regex", "regex": "^\\d+$"}],
+    ["--timesketch_endpoint", "Timesketch endpoint", null],
```
jleaniz, line 45: Should we set this to the default Timesketch URL? ```http://127.0.0.1:5000``` or will this be read from the ```data/config.json``` file?

dftimewolf/lib/exporters/timesketch.py
```diff
@@ -53,10 +53,13 @@ def __init__(
...
       self,
-      incident_id: None = None,
-      sketch_id: Optional[int] = 0,
-      analyzers: None = None,
-      token_password: str = '',
+      incident_id: str,
+      sketch_id: Optional[int],
+      analyzers: str,
+      token_password: str,
+      endpoint: Optional[str],
```
jleaniz, line 60: Optional seems weird, since you will never be able to connect without this being set somehow?

jleaniz (dismissed):
LGTM. just a couple comments.

data/recipes/workspace_user_login_ts.json
tomchop: It's very likely to be replaced by *some* config file, but we can set a default.

dftimewolf/lib/exporters/timesketch.py
tomchop: It's optional because `token_password` implies that you already have the endpoint and everything stored in the token.

tomchop:
merging as @jleaniz already approved

PR #799, opened 2023-11-15
Fix jenkins build

tomchop:
* Removed the token usage in favor of hardcoded but explicit credentials set in the jenkins runner
* Reorganized config directory
* Added the test runner config to version control


jkppr (approved):
looks good

PR #824, opened 2024-01-11
Fix: Use Timesketch config generator if no token password or credentials are provided

tomchop:


jkppr (approved):
lgtm, thanks for fixing the issue.

PR #843, opened 2024-03-11
Update main_tool.py

sa3eed3ed:
make sure only required args exist in recipe, avoid adding an optional arg to all other recipes that would pass it as None


sa3eed3ed:
as discussed will close this to keep all default values documented in recipes

PR #856, opened 2024-05-07
Add new container types, tweak GRR Artifact Collector to take them into account

tomchop:
* GRRArtifact containers, so other modules can instruct the GRR collectors what to collect
* SQLQuery containers, so that other modules can instruct what queries to run in other collectors.

### DIFF
--- a/dftimewolf/lib/containers/containers.py
+++ b/dftimewolf/lib/containers/containers.py
@@ -673,8 +673,8 @@ class SQLQuery(interface.AttributeContainer):
   """SQL Query container.
 
   Attributes:
-    query (str): The SQL query string..
-    description (str): A description of the query.
+    query: The SQL query string.
+    description: A description of the query.
   """
 
   CONTAINER_TYPE = "sql_query"
@@ -753,7 +753,7 @@ class GRRArtifact(interface.AttributeContainer):
   """GRR Artifact container.
 
   Attributes:
-    name (str): Name of the GRR artifact.
+    name: Name of the GRR artifact.
   """
 
   CONTAINER_TYPE = "grr_artifact"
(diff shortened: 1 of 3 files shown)


dftimewolf/lib/containers/containers.py
```diff
@@ -669,6 +669,30 @@ def __str__(self) -> str:
...
+
+  Attributes:
+    query: The SQL query string.
+    description: A description of the query.
+  """
+
+  CONTAINER_TYPE = "sql_query"
+
+  def __init__(self, query: str, description: str) -> None:
+    super(SQLQuery, self).__init__()
```
jleaniz, line 683:
```suggestion
    super().__init__()
```
Python 3 doesn't require explicit arguments.

dftimewolf/lib/containers/containers.py
```diff
@@ -723,3 +747,24 @@ def __str__(self) -> Optional[str]:
...
+
+  Attributes:
+    name: Name of the GRR artifact.
+  """
+
+  CONTAINER_TYPE = "grr_artifact"
+
+  def __init__(self, name: str):
+    self.name = name
+
```
jleaniz, line 763:
```suggestion
super().__init__()

```

jleaniz (changes requested):
Left two small suggestions, which i think will also fix the pylint warning. LGTM otherwise.

dftimewolf/lib/containers/containers.py
```diff
@@ -723,3 +747,24 @@ def __str__(self) -> Optional[str]:
...
+  """GRR Artifact container.
+
+  Attributes:
+    name: Name of the GRR artifact.
+  """
+
+  CONTAINER_TYPE = "grr_artifact"
+
+  def __init__(self, name: str):
+    self.name = name
```
jleaniz, line 762:
```suggestion
        super().__init__()
        self.name = name
```

jleaniz (changes requested):

PR #857, opened 2024-05-08
Account for payload provided by CollectBrowserHistory

tomchop:
`CollectBrowserHistory` flow's payload are structured slightly differently than ArtifactCollectorFlow payloads or TimelineFlow payloads, so we account for that here to process files.


jkppr (approved):
lgtm

PR #861, opened 2024-05-14
Collect turbinia processor telemetry

jleaniz:


dftimewolf/lib/utils.py
```diff
@@ -17,6 +18,11 @@
 TOKEN_REGEX = re.compile(r'\@([\w_]+)')
 
 
+def CalculateRunTime(time_start: float) -> float:
+  """Calculates a time delta used for runtime calulcations."""
```
tomchop, line 22: Given all the operations happening, it would be nice to specify which units we're taking in and returnin (e.g. seconds, milliseconds, etc.)

dftimewolf/lib/state.py
```diff
@@ -389,8 +394,11 @@ def _RunModuleProcessThreaded(
             f'Launching {module.name}.Process thread with {str(c)} from '
             f'{c.metadata.get(interface.METADATA_KEY_SOURCE_MODULE, "Unknown")}'
         )
+        time_start = time.time()
         futures.append(executor.submit(module.Process, c))
-
+        total_time = utils.CalculateRunTime(time_start)
+        TELEMETRY.LogTelemetry(
+          'total_time', str(total_time), 'core', self.recipe.get('name', 'no_name'))
```
tomchop, line 401: I'd argue that the key here should rather be something like `total_time_<module_name>` so that we can differentiate the entries per module.

Alternatively - you can also change the `'core'` parameter to be `core-<module_name>`.

tomchop: Same comment for the changes in L365

dftimewolf/lib/processors/turbinia_gcp.py
```diff
@@ -241,13 +254,33 @@ def Process(self, request_container: containers.TurbiniaRequest) -> None:
...
+        method_name = profiler_entry.code
+      else:
+        method_name = profiler_entry.code.co_name
+      if method_name in self.profiler_methods:
+        telemetry_entry[method_name] = (
+            f'callcount: {str(profiler_entry.callcount)}, '
+            f'tottime :{str(round(profiler_entry.totaltime * 1000, 10))},'
+            f'inlinetime: {str(round(profiler_entry.inlinetime * 1000, 10))}'
+        )
+    self.LogTelemetry(telemetry_entry)
```
tomchop, line 282: For readbility, consider extracting this chunk to its own function?

tomchop (changes requested):

tomchop (approved):

PR #878, opened 2024-05-30
Remove apply_parsers args

tomchop:


ramo-j (approved):
Noting the failing pylint check, but otherwise LGTM.

PR #879, opened 2024-06-05
Refactored validators into separate files

Onager:
For ease of viewing and isolating


dftimewolf/lib/validators/manager.py
```diff
@@ -0,0 +1,118 @@
...
+    """Returns a list of all registered validators.
+
+    Returns:
+      A list of all registered validators.
+    """
+    return list(cls._validator_classes.keys())
+
+
+  @classmethod
+  def RegisterValidator(
```
tomchop, line 24: Other manages have a boolean parameter that allows for overwriting validators (useful if we have different internal validation logic)

tomchop (changes requested):
a few questions, otherwise LGTM!

dftimewolf/lib/validators/gcp_zone.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+class GCPZoneValidator(AbstractValidator):
+  """Validates a GCP zone."""
+
+  # Source: https://cloud.google.com/compute/docs/regions-zones/
+  # Fetched 2023-01-13
+  # TODO - Fetch at runtime?
+  _zones = {
```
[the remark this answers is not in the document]
Onager: Ack, done.

dftimewolf/lib/validators/manager.py
Onager: Added an override argument here

tomchop (approved):

PR #880, opened 2024-06-05
Change datetime validator to produce datetime objects rather than strings

Onager:
-  Modules expecting datetime arguments have been updated to accept datetime objects
- The sanity_check preflight has been removed, as it was unused and duplicates functionality


data/recipes/aws_logging_collect.json
```diff
@@ -24,7 +24,7 @@
         ["region", "AWS Region", null, {"format": "aws_region"}],
         ["--profile_name", "Name of the AWS profile to collect logs from.", null],
         ["--query_filter", "Filter expression to use to query logs.", null],
-        ["--start_time", "Start time for the query.", null, {"format": "datetime", "format_string": "%Y-%m-%d %H:%M:%S", "before": "@end_time"}],
-        ["--end_time", "End time for the query.", null, {"format": "datetime", "format_string": "%Y-%m-%d %H:%M:%S", "after": "@start_time"}]
+        ["--start_time", "Start time for the query.", null, {"format": "datetime", "before": "@end_time"}],
+        ["--end_time", "End time for the query.", null, {"format": "end_time",  "after": "@start_time"}]
```
tomchop, line 28: `datetime_end` might be a better choice here (and elsewhere)

tomchop (dismissed):
one comment, otherwise LGTM

data/recipes/aws_logging_collect.json
Onager: Yes, you're right, updated

PR #882, opened 2024-06-07
Fix optional argument handling

Onager:


dftimewolf/cli/dftimewolf_recipes.py
```diff
@@ -325,10 +325,9 @@ def ValidateArguments(self) -> None:
 
       switch = expanded_argument.switch.replace('--', '')
       argument_mandatory = switch == arg.switch
-      argument_set = switch in self.state.command_line_options
+      argument_value = self.state.command_line_options[switch]
```
tomchop, line 328: Maybe use .get() here to avoid key errors?

tomchop (dismissed):
One small nit

dftimewolf/cli/dftimewolf_recipes.py
Onager: Yes, definitely

PR #890, opened 2024-06-21
Fix error when handling turbinia_auth flag

jleaniz:
Fixes an error handling the ```turbinia_auth``` flag.


dftimewolf/lib/processors/turbinia_base.py
```diff
@@ -385,6 +390,15 @@ def TurbiniaSetUp(
     """
     self.project = project
     self.turbinia_auth = turbinia_auth
+    # turbinia_auth is a boolean in the recipe but gets passed as a string
+    # to the module, so we check for the correct type here.
```
tomchop, line 394: AFAIR recipe args defined with `bool` default values are passed on as Booleans to the SetUp function. Maybe this changed with the recent arg validator changes. But modules shouldn't have to cast it

tomchop (changes requested):

dftimewolf/lib/processors/turbinia_base.py
Onager: This was a bug in the validator changes, which should be fixed by #894

tomchop (approved):

PR #891, opened 2024-07-03
Refactor GRR file downloads

ramo-j:
* Updated file downloads to always use Blob download, as archive downloads are not always available and determining whether archives are available is complex (An update to the GRR api to allow tar.gz archives seems like it might fix this)
* Removed the custom `_AwaitFlow` method and replaced it with the GRR native `WaitUntilDone`


sydp (approved):
LGTM.

non-blocking side note, might be worth a clean up of single/double quoting in a future PR.

PR #894, opened 2024-07-09
Fixes for issues relating to validation changes

Onager:


data/recipes/gcp_logging_cloudsql_ts.json
```diff
@@ -17,7 +17,9 @@
         "project_name": "@project_name",
         "filter_expression": "logName:\"projects/@project_name/logs/cloudsql.googleapis.com\" timestamp>\"@start_date\" timestamp<\"@end_date\"",
```
tomchop, line 18: Do we want to do the interpolation here as well? (and in other recipes?)

tomchop (changes requested):

tomchop (approved):

PR #898, opened 2024-07-22
Expand Recipe testing to perform arg validation

ramo-j:


data/recipes/aws_forensics.json
```diff
@@ -26,7 +27,7 @@
       ["--instance_id", "Instance ID of the instance to analyze.", null, {"format": "regex", "comma_separated": false, "regex": "^i-[0-9a-f]{8,17}$"}],
       ["--volume_ids", "Comma-separated list of volume IDs to copy.", null, {"format": "regex", "comma_separated": true, "regex": "^vol-[0-9a-f]{8,17}$"}],
       ["--all_volumes", "Copy all volumes in the designated instance. Overrides volume_ids if specified.", false],
-      ["--boot_volume_size", "The size of the analysis VM boot volume (in GB).", 50, {"format": "regex", "regex": "^\\d+$"}],
+      ["--boot_volume_size", "The size of the analysis VM boot volume (in GB).", "50", {"format": "regex", "regex": "^\\d+$"}],
```
Onager, line 30: This isn't actually a string though, it's a number - we should add a new validator for numbers

ramo-j: Added an integer parser.

tests/cli/main_tool.py
```diff
@@ -129,11 +133,22 @@ def testRecipeSetupArgs(self, recipe_name):
...
     # pylint: disable=protected-access
     self.tool._state = dftw_state.DFTimewolfState(config.Config)
     recipe = self.tool._recipes_manager.Recipes()[recipe_name]
 
+    test_params = recipe.GetTestParams()
+    if test_params:
+      recipe_args = [recipe_name] + test_params
+      self.tool.ParseArguments(recipe_args)
+    else:
+      self.fail('No test_params in recipe')
```
tomchop, line 150: This essentially makes the test_params mandatory for all recipes, even those that may not have any (as unlikely a it sounds it is something that is supported). I'm not against it, but wanted to make sure that was on your radar.

dftimewolf/lib/resources.py
```diff
@@ -62,3 +62,9 @@ def GetHelpString(self) -> str:
     short_description = self.contents.get(
         'short_description', 'No description')
     return ' {0:<35s}{1:s}\n'.format(self.name, short_description)
+
+  def GetTestParams(self) -> Optional[list[str]]:
+    """Get the test params from a recipe."""
+    if self.contents.get('test_params', None):
+      return str(self.contents.get('test_params', '')).split(' ')
+    return None
```
tomchop, line 70: We could also always return an empty list if no test_params are found (the same truthiness as `None`)

tomchop (dismissed):
Some small nits, but otherwise LGTM! I'll let you adjust if you feel like it and merge.

dftimewolf/lib/resources.py
ramo-j: Related to the other comment, changed up to recognise the difference between an empty test arg list `[]` and the recipe missing the `test_params` field (now throws exception in the latter case.)

tests/cli/main_tool.py
ramo-j: Having `test_params` required for every recipe was indeed the intent, but the no args edge case didn't occur to me. Updated, with test cases.

tomchop (approved):

PR #904, opened 2024-08-14
Add new flag to filter findings in Turbinia reports according to  the finding priority. Pump turbinia-client to 1.0.7 and fix some mypy and linter findings

sa3eed3ed:
add new flag to filter findings in Turbinia reports according to  the finding priority


hacktobeer (commented):
LGTM for the Turbinia functional side.
@tomchop can you approve from the dftimewolf side?

tomchop (approved):

PR #905, opened 2024-08-15
Do not force raw access depending on platform

tomchop:


jleaniz (approved):
LGTM although i don't know why the change is happening :)

PR #925, opened 2024-10-18
New GC Recipe gcp_logging_l2t_ts

tmoney-coder:
Added a new recipe that will collect the logs from GC, and then parse them with Docker Plaso, and then submit to Timesketch.

The current gcp_logging_ts doesn't process the data via Plaso. As such, the data is in a different format then with other gcp log collection options. This resolves that issue.

### DIFF
--- a/data/recipes/gcp_logging_l2t_ts.json
+++ b/data/recipes/gcp_logging_l2t_ts.json
@@ -1,0 +1,64 @@
+{
+  "name": "gcp_logging_l2t_ts",
+  "short_description": "Collects logs from a GCP project and sends them to Timesketch.",
+  "description": "Collects logs from a GCP project, parses the logs with Plaso, and sends them to Timesketch. https://cloud.google.com/logging/docs/view/query-library for example queries.",
+  "test_params": "project-name filter_expression",
+  "preflights": [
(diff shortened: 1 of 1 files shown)


ramo-j:
This new recipe is very similar to `data/recipes/gcp_logging_cloudaudit_ts.json` - The only difference I can see is that there is a fixed query in the existing, and no fixed or default query in the new. It is probably better to update the existing to allow the user to specify a query, with the one that is currently hardcoded in used as a default if the user provides no query. WDYT?

tmoney-coder:
@ramo-j I moved it to a new file as there could be folks that use the existing format for their workflows and I didn't want to intrudce an error existing users would encounter. I am more then happy to update the existing flow.

ramo-j:
> @ramo-j I moved it to a new file as there could be folks that use the existing format for their workflows and I didn't want to intrudce an error existing users would encounter. I am more then happy to update the existing flow.

If the previous fixed query is provided as a default, then existing workflows shouldn't be affected. Updating the existing would be a better way to go I think.

tmoney-coder:
Moved back to original recipe and then tested working as of 11/20/24

PR #931, opened 2024-11-11
Added a Dataframe to Filesystem exporter

ramo-j:


dftimewolf/lib/exporters/df_to_filesystem.py
```diff
@@ -0,0 +1,176 @@
...
+    with open(output_path, 'w') as f:
+      container.data_frame.to_json(
+        f, orient='records', lines=True, default_handler=str)
+
+    self.state.StoreContainer(container=containers.File(
+        name=os.path.basename(output_path),
+        path=output_path,
+        description=container.description))
+
+    self.logger.debug(f'Export of {container.name} to {output_path} complete')
```
tomchop, line 141: This function is nearly identical to `_ExportDataFrameCSV` and `_ExportDataFrameMarkdown`, except for the pandas `to_` function being called. Can we factor that code into `_ExportSingleDataFrame` and to the switch case around the `to_` function?

tomchop (changes requested):

dftimewolf/lib/exporters/df_to_filesystem.py
ramo-j: Yeah, good call. Updated, PTAL.

tomchop (approved):

PR #936, opened 2024-12-04
Added attributes types and new attribute status_reasons

roshanmaskey:
Changes
- Added attribute definition in `GCPLoggingTimesketch`
- Changed `dcsa_email: str` to `dcsa_emails: list[str]`
- Added `status_reasons: list[str]`
- Added unit test `testServiceAccountCreateFailed`

### DIFF
--- a/dftimewolf/lib/processors/gcp_logging_timesketch.py
+++ b/dftimewolf/lib/processors/gcp_logging_timesketch.py
@@ -5,7 +5,7 @@ import json
 import re
 import tempfile
 from datetime import datetime
-from typing import Any, Dict, Optional, TYPE_CHECKING
+from typing import Any, Dict, List, Optional, TYPE_CHECKING
 
 from dftimewolf.lib.containers import containers
 from dftimewolf.lib.module import BaseModule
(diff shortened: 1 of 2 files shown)


dftimewolf/lib/processors/gcp_logging_timesketch.py
```diff
@@ -16,7 +16,47 @@
...
+        Engine instance.
+    source_ranges (List[str]): firewall source ranges.
+    status_code (str): operation success or failure code.
+    status_message (str): operation success or failure message.
+    status_reason (str): operation failure reasons.
+    textPayload (str): text payload for logs not using a JSON or proto payload.
+    timestamp_desc (str): description of timestamp.
+    user (str): user or requestor.
+    user_agent (str): user agent used in the request.
+  """
```
ramo-j, line 59: The `Attributes` list in a class docstring usually refers to the members of the class, but this appears to be log fields in the output. Maybe it would be better to list these as a file level comment/docstring?

roshanmaskey: Changed to file docstring.

ramo-j (approved):
LGTM

PR #943, opened 2024-12-18
Create OpenRelik processor module

jleaniz:
This PR adds a new OpenRelik processor.


dftimewolf/lib/processors/openrelik.py
```diff
@@ -0,0 +1,174 @@
...
+  @staticmethod
+  def GetThreadOnContainerType() -> Type[interface.AttributeContainer]:
+    return containers.File
+
+  def GetThreadPoolSize(self) -> int:
+    return 3
+
+  @staticmethod
+  def KeepThreadedContainersInState() -> bool:
+    return False
```
tomchop, line 165: We don't need these

dftimewolf/lib/processors/openrelik.py
```diff
@@ -0,0 +1,174 @@
...
+  def PollWorkflowStatus(self, workflow_id: int) -> Iterator[str]:
+    """Polls the status of a workflow until it completes."""
+    filename = str(workflow_id)
+    workflow = self.openrelik_workflow_client.get_workflow(
+      self.folder_id, workflow_id
+    )
+    status = workflow.get("tasks")[0].get("status_short")
+    output_file_ids = []
+    while status != "SUCCESS" and status != "FAILED":
+      self.logger.info(f"Workflow {workflow_id} status: {status}")
```
tomchop, line 83: Let's be a bit more conservative in what we're logging and move some of this to debug (especially since it can pop up many times)

dftimewolf/lib/processors/openrelik.py
```diff
@@ -0,0 +1,174 @@
...
+      status = workflow.get("tasks")[0].get("status_short")
+    self.logger.info(f"Workflow {workflow_id} status: {status}")
+    if status == "FAILED":
+      self.ModuleError(f"Workflow {workflow_id} failed", critical=True)
+
+    for task in workflow.get("tasks"):
+      output_files = task.get("output_files", [])
+      for output_file in output_files:
+        output_file_id = output_file.get("id")
+        filename = output_file.get("display_name", workflow_id)
```
tomchop, line 97: This gets overwritten every time the loop goes, and in L101 you only get the "last" filename. WAI?

dftimewolf/lib/processors/openrelik.py
```diff
@@ -0,0 +1,174 @@
...
+    self.incident_id = incident_id
+    self.sketch_id = sketch_id
+
+  def PollWorkflowStatus(self, workflow_id: int) -> Iterator[str]:
+    """Polls the status of a workflow until it completes."""
+    filename = str(workflow_id)
+    workflow = self.openrelik_workflow_client.get_workflow(
+      self.folder_id, workflow_id
+    )
+    status = workflow.get("tasks")[0].get("status_short")
```
tomchop, line 80: Calling `get` here might return None, which will raise an exception when you try to access `[0]`. Can we check the value of "tasks" returns a thing that is not empty / not None?

dftimewolf/lib/processors/openrelik.py
```diff
@@ -0,0 +1,174 @@
...
+    """Downloads a file from OpenRelik.
+
+    Args:
+        file_id: The ID of the file to download.
+        filename: The name of the file to download.
+
+    Returns:
+        str: The path to the downloaded file.
+    """
+    endpoint = f"{self.openrelik_api_client.base_url}/files/{file_id}/download"
```
tomchop, line 114: Maybe add a logging.info call to indicate that we're downloading a file? since it can take a long time, we don't want users to think this is stuck

tomchop (changes requested):

dftimewolf/lib/processors/openrelik.py
jleaniz: Done

dftimewolf/lib/processors/openrelik.py
jleaniz: Done

dftimewolf/lib/processors/openrelik.py
jleaniz: Updated this to use a dictionary with the output_file_id and filename instead.

dftimewolf/lib/processors/openrelik.py
jleaniz: We do for now since this is still a ```ThreadAwareModule````

dftimewolf/lib/processors/openrelik.py
```diff
@@ -0,0 +1,181 @@
...
+    self.workflow_id: int | None = None
+    self.folder_id: int | None = None
+    self.sketch_id: int | None = None
+    self.incident_id: str | None = None
+
+  # pylint: disable=arguments-differ
+  def SetUp(
+    self,
+    incident_id: str | None,
+    sketch_id: int | None,
```
tomchop, line 51: Looks like this is never used?

tomchop (dismissed):
one small nit, otherwise lgtm!

PR #951, opened 2025-01-14
Additional module testing refactoring.

ramo-j:
Creates a method in the module testing base class for consistently replicating how modules are run by the orchestration in testing.


sydp (approved):
Lgtm

PR #952, opened 2025-01-14
Add TimesketchEvents container to containers.py

itsmvd:


dftimewolf/lib/containers/containers.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import datetime
+import pandas as pd
```
tomchop, line 7: Remove import from the TYPE_CHECKING block

tomchop (approved):

dftimewolf/lib/containers/containers.py
itsmvd: Done

tomchop (approved):

PR #959, opened 2025-01-21
remove start_datetime and end_datetime as required args, fix bug limiting amount of searches to 10k events

itsmvd:


sydp (approved):
Lgtm

PR #966, opened 2025-02-06
Container manager bugfix

ramo-j:
Correct an issue where the container manager popped incorrect containers; tests for same.

### DIFF
--- a/dftimewolf/lib/containers/manager.py
+++ b/dftimewolf/lib/containers/manager.py
@@ -84,6 +84,7 @@ class ContainerManager():
   def GetContainers(self,
                     requesting_module: str,
                     container_class: Type[interface.AttributeContainer],
+                    pop: bool = False,
                     metadata_filter_key: str | None = None,
                     metadata_filter_value: Any = None
       ) -> Sequence[interface.AttributeContainer]:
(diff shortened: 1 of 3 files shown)


dftimewolf/lib/containers/manager.py
```diff
@@ -127,9 +127,11 @@ def GetContainers(self,
...
       if pop:
         # A module can only pop containers it has stored.
-        for c in ret_val:
-          if c in self._modules[requesting_module].storage:
-            self._modules[requesting_module].storage.remove(c)
+        # Remove by unique object id: Not __eq__() in case there are dupes, or
+        #   attempting to compare different types of containers.
+        self._modules[requesting_module].storage = [
+            c for c in self._modules[requesting_module].storage
+            if id(c) not in [id(c) for c in ret_val]]
```
sydp, line 134: `[id(c) for c in ret_val]` looks to me like it remains constant through out the outer loop/comprehension - could you store this as a variable before?  It would avoid having nested comprehensions which can be difficult to read.

If not, I would recommend choosing a different name in the inner comprehension (`c`) to avoid variable shadowing.

sydp (changes requested):

dftimewolf/lib/containers/manager.py
ramo-j: Done.

PR #967, opened 2025-02-07
Add source module information to container discrepancy telemetry

ramo-j:


Fryyyyy (approved):
I'm assuming the receiving end will also be changed to accept a list of dicts rather than a list of strings

PR #976, opened 2025-03-07
RIP in peace, metawolf.

ramo-j:


giovannt0 (approved):
rip in peace little wolf

PR #978, opened 2025-03-11
Add GRR MPA measurements to telemetry

tomchop:


dftimewolf/lib/collectors/grr_base.py
```diff
@@ -85,6 +85,7 @@ def _WrapGRRRequestWithApproval(
       grr_object: Union[Hunt, Client],
       grr_function: Callable,  # type: ignore[type-arg]
       logger: Logger,
+      telemetry: Callable[[dict[str, str]], None],
```
ramo-j, line 88: Nit - maybe `log_telemetry`, or `telemetry_callback` for the name?

tests/lib/collectors/grr_base.py
```diff
@@ -86,16 +86,19 @@ def testApprovalWrapper(self, _):
...
-        mock_forbidden_function,
-        logging.getLogger('GRRBaseModuleTest'),
-        'random1',
-        'random2',
-        random3=4,
-        random4=4)
+      mock_grr_object,
+      mock_forbidden_function,
+      logging.getLogger("GRRBaseModuleTest"),
+      mock_telemetry,
```
ramo-j, line 96: Is it worth adding an assertion that telemetry gets called how you expect? (Or, possibly even adding a dedicated test for that.)

ramo-j (approved):
Approved with some small notes

PR #980, opened 2025-03-17
Remvoed TimesketchEnhancer.

ramo-j:
Also, formatted every recipe as a side effect of using `jq` for bulk updates.


tomchop (approved):
Approved, although I think it might be useful to keep the module while we make sure we don't need it (in case we need to bring it back)

ramo-j:
> Approved, although I think it might be useful to keep the module while we make sure we don't need it (in case we need to bring it back)

Eh, if we need to bring it back, it will always be in source control history, and our hearts.

PR #981, opened 2025-03-17
Make TS collector better at guessing sketches to use

tomchop:


tests/lib/collectors/timesketch.py
ramo-j, line 1: Can you add a test that checks the precedence of provided IDs? That is, if there is a sketch id passed in as a param to SetUp, a different one in the cache, and a different one in a container, then the one passed in to SetUp is the one used. (And possibly other combinations where two IDs are provided.)

ramo-j (changes requested):
Just noting - Review requested, but this is still marked draft?

Also, linter appeasement required.

ramo-j (approved):

PR #985, opened 2025-04-10
Make OpenRelik module errors non critical for now

jleaniz:
Temporarily make module errors non critical while the module is tested out.

### DIFF
--- a/tests/lib/processors/openrelik.py
+++ b/tests/lib/processors/openrelik.py
@@ -5,7 +5,7 @@ from unittest import mock
 
 from openrelik_api_client import api_client, folders, workflows
 
-from dftimewolf.lib import errors
+# from dftimewolf.lib import errors
 from dftimewolf.lib.containers import containers
 from dftimewolf.lib.processors import openrelik as openrelik_processor
 from tests.lib import modules_test_base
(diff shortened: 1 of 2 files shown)


tests/lib/processors/openrelik.py
```diff
@@ -42,8 +42,11 @@ def testPollWorkflowStatus(self, mock_download_workflow, mock_get_workflow):
     self._module.openrelik_workflow_client = workflows.WorkflowsAPI(
       api_client.APIClient("fake_api", "fake_key")
     )
-    status_generator = self._module.PollWorkflowStatus(456)
-    self.assertRaises(errors.DFTimewolfError, next, status_generator)
+    # Commenting this out since now the module won't raise a ciritical error
+    # Uncomment these two lines when critical errors are re-enabled.
+
+    # status_generator = self._module.PollWorkflowStatus(456)
```
ramo-j, line 48: Rather than comment it out, what we want at the moment is to test that no error is raised.

ramo-j (approved):
Approved with one note

tests/lib/processors/openrelik.py
jleaniz: Done.

PR #1019, opened 2025-08-25
Update GRR client approval links - part 2

s-westphal:


ramo-j (commented):
Just for doco purposes, can you please point to a link upstream in GRR indicating this change is required? Thanks!

PR #1063, opened 2026-04-28
Logging improvements

ramo-j:
This changes the per-module logging handlers, such that modules correctly use the logging handlers configured by the orchestration.


dftimewolf/lib/modules/module_runner.py
```diff
@@ -233,6 +226,7 @@ def _InvokeModulesInThreads(self, callback: typing.Callable[[typing.Any], typing
     for module_definition in self._recipe['modules']:
       thread_args = (module_definition,)
       thread = threading.Thread(target=callback, args=thread_args)
+      thread.name = thread.name[:thread.name.find(' ')]
```
sydp: if there is no space in thread.name it seems this truncates, is that intended?  
perhaps it might be more readable to just use thread.name.split(' ')[0].

dftimewolf/lib/logging_utils.py
```diff
@@ -113,17 +75,18 @@ def format(self, record: LogRecord) -> str:
...
-      stack = [i.function for i in inspect.stack()]
-      if 'Process' in stack:
-        thread_name = threading.current_thread().name
-        message = str(record.msg)
-        record.msg = f"[{thread_name}] {message}"
-    return super(WolfFormatter, self).format(record)
+        formatted_message = loglevel_color + formatted_message + RESET_SEQ
+
+    record.msg = formatted_message
+
```
sydp: any reason these new lines have extra empty lines between them?

sydp (approved):
some minor questions but otherwise lgtm

dftimewolf/lib/modules/module_runner.py
ramo-j: Updated.

dftimewolf/lib/logging_utils.py
ramo-j: Not a good one. Updated.

PR #1089, opened 2026-07-24
Add opentelemetry instrumentation

aarontp:
Introduces new opentelemetry.py and renames telemetry.py and related data to spanner_telemetry.py.  Updates and separates related config vars as well.

Adds instrumentation for module execution and calls to external services.


ramo-j (approved):
LGTM but you probably need a dependency on `opentelemetry-api` in pyproject.toml. Similar to how the current spanner one is optional, this should probably be an optional dependency group also.

aarontp:
> LGTM but you probably need a dependency on `opentelemetry-api` in pyproject.toml. Similar to how the current spanner one is optional, this should probably be an optional dependency group also.

Done.  Thanks!  Merging.

PR #1093, opened 2026-08-14
Remove mypy and pytype in favour of pyrefly

ramo-j:
Remove mypy and pytype github workflows in favour of a pyrefly workflow. Also, updates to suit pyrefly and removing pytype / mypy annotations so that check passes.

Reasoning: 
* pytype is not supported any more
* mypy and pyrefly have a lot of overlap


tomchop (approved):
Good change! Thanks

PR #1097, opened 2026-08-25
Update duplicate container comparison when used for storing dupe containers into the container manager

ramo-j:


dftimewolf/lib/containers/manager.py
```diff
@@ -127,6 +127,10 @@ def StoreContainer(self,
           else:
             if container.CONTAINER_TYPE not in module.storage:
               module.storage[container.CONTAINER_TYPE] = []
+
+            # If the container to add exists already in the state, don't add it again
+            if id(container) in [id(c) for c, _ in module.storage[container.CONTAINER_TYPE]]:
+              continue
```
Onager, line 134: A debug log here might be useful, to track where/why this is happening. 

Slight readability/performance thing: try `if any(c is container for c, _ in module.storage[...`  

* `is` also checks memory addresses but is a bit faster, 
* any() uses a generator rather than building a whole new list of ids

Onager (changes requested):

dftimewolf/lib/containers/manager.py
ramo-j: Updated, PTAL.

tests/lib/containers/manager.py
```diff
@@ -548,43 +548,38 @@ def test_StoreDuplicateContainers(self):
...
-    c6 = _TestContainer3('param1')
-    c5.SetMetadata('key', 'foo')
-    c6.SetMetadata('key', 'bar')
-    self._container_manager.StoreContainer(
-        source_module='Preflight1', container=c5)
-    self._container_manager.StoreContainer(
-        source_module='Preflight1', container=c6)
+        source_module='Preflight1', container=container_3)
 
     # Dataframe members of containers have special handling; check that too
```
Onager: What's special about them? I don't understand what this is testing

Onager (approved):

tests/lib/containers/manager.py
ramo-j: In the old logic, the internal dataframes needed to be compared between containers. This specific test is probably obsolete now, come to think of it.

